PR #2725, opened 2020-05-03
Improve transport::Publisher reliability

nlamprian:
It has been observed that `OnPublishComplete` can run before `result` is checked in `SendMessage`. In that case, the `pubIds` entry is deleted by `OnPublishComplete`, and then it's recreated in `SendMessage`. This [blocks publication of any future messages](https://github.com/osrf/gazebo/blob/gazebo9/gazebo/transport/Publisher.cc#L165), thus rendering the publisher unusable.

The fix guards access to `pubIds` in `SendMessage`, and it processes the `pubIds` entry only if it still exists.

### DIFF
--- a/test/regression/2724_failing_publisher.cc
+++ b/test/regression/2724_failing_publisher.cc
@@ -1,0 +1,108 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#ifndef _WIN32
+#include <unistd.h>
+#endif
+#include "gazebo/test/ServerFixture.hh"
+
+using namespace gazebo;
+using namespace gazebo::transport;
+
+class FailingPublisherTest : public ServerFixture
+{
+};
+
+class TestTransport : public CallbackHelper
+{
+  /////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


chapulina:
Thanks for the PR, @nlamprian . Do you think it would be possible to write a test that exercises the fix?

nlamprian:
Oof, this will require some effort. I'll need to familiarize myself with the full implementation to understand how to trigger the conditions. I'll give it a try.

nlamprian:
The unit test is added. It successfully reproduces the error. Basically, for the error to show up, there have to be callbacks in the `Publication`. A callback is added by a subscription to the `Publication`, which, in this case, happens in the `ConnectionManager` with a `SubscriptionTransport`. It's still not clear to me how everything around all of this works, but nonetheless, I replicated the conditions in the test. There, it can be seen that the first message is sent and received, but the second one is not.

nlamprian:
May I please ask for an update on this? Is there something holding back the review? The publisher continues to be a problem for me, and I would hope to see it fixed soon.

chapulina:
@osrf-jenkins run tests

Thank you for the contribution and the test, it looks good to me. Let's just run a round of CI to make sure everything is ok.

chapulina:
@osrf-jenkins run tests again?

test/regression/2724_failing_publisher.cc
```diff
@@ -0,0 +1,108 @@
...
+  ++g_receivedMsgs;
+}
+
+/////////////////////////////////////////////////
+TEST_F(FailingPublisherTest, DirectPublish)
+{
+  Load("worlds/empty.world");
+
+  // Initialize node
+  NodePtr node = boost::make_shared<Node>();
```
chapulina: This is failing compilation on CI:

```
/var/lib/jenkins/workspace/gazebo-ci-pr_any-ubuntu_auto-amd64-gpu-nvidia/gazebo/test/regression/2724_failing_publisher.cc:67:18: error: ���make_shared��� is not a member of ���boost���
   NodePtr node = boost::make_shared<Node>();
```

I think you may need to use `new` with the supported boost version.

chapulina (changes requested):

nlamprian:
@osrf-jenkins run tests

chapulina:
@osrf-jenkins run tests

chapulina (approved):
The fix and test look good and CI is happy :+1:

Issue #2729, opened 2020-05-10
Depth Camera Sensor Scoped Name Incorrect Gazebo 9


blakermchale:
I am currently using a custom realsense gazebo plugin on my robot in gazebo and trying to simulate multiple vehicles. The custom gazebo plugin instantiates 4 cameras, 3 of which are normal cameras while the last one is a depth camera. When the sensors load I use gzerr to output the scoped name of the sensor after grabbing it through the sensor manager. The scoped names for the normal cameras are `default::drone_0::realsense_camera::base_link::color` which is correct. However, the depth cameras scoped name is `default::default::depth` when it should be `default::drone_0::realsense_camera::base_link::depth`. Currently, it is not inheriting its parents names from the sdf.

Is this a bug in gazebo or is there something I am doing incorrect? A sample of the depth camera in the sdf code is below. I am using sdf version 1.5.

```
<?xml version="1.0"?>
<sdf version="1.5">
<model name="realsense_camera">
    <plugin name="realsense_plugin" filename="librealsense_gazebo_plugin.so"/>
    <pose frame="">0 0 0.015 0 0 0</pose>
    <link name="base_link">
      ...
      ...
      <sensor name="depth" type="depth">
        <pose frame="">0 -0.03 0.004 0 0 0</pose>
        <camera name="__default__">
          <horizontal_fov>1.047</horizontal_fov>
          <image>
            <width>640</width>
            <height>480</height>
          </image>
          <clip>
            <near>0.1</near>
            <far>100</far>
          </clip>
          <noise>
            <type>gaussian</type>
            <mean>0.0</mean>
            <stddev>0.007</stddev>
          </noise>
        </camera>
        <always_on>1</always_on>
        <update_rate>60</update_rate>
        <visualize>0</visualize>
      </sensor>
    </link>
  </model>
</sdf>
```

blakermchale:
I found where the issue occurs in the code. It occurs on line 78-79 of DepthCameraSensor.cc:
```
this->dataPtr->depthCamera = this->scene->CreateDepthCamera(
        this->sdf->Get<std::string>("name"), false);
```
I believe it should be like lines 130-131 of CameraSensor.cc:
```
std::string scopedName = this->parentName + "::" + this->Name();
    this->dataPtr->depthCamera = this->scene->CreateDepthCamera(scopedName, false);
```
The code on lines 99-106 in DepthCameraSensor.cc would also need to be updated to have their this->Name() replaced by scopedName.

[...]

PR #2730, opened 2020-05-11
Fix DepthCameraSensor Scoped Name

blakermchale:
Currently, the Depth Camera Sensor does not create its sensor with a scoped name like the Camera Sensor. This is an issue because it becomes difficult maintaining multiple vehicles with the same depth sensor name. Currently, there is no way to access a sensor unique to a specific model but with the same name as other models sensors. This pull request should fix this issue. It should be noted that this problem persists in gazebo10 and gazebo11. I created issue #2729 that describes the problem more.


iche033:
Changes look good. The main concern is that it may break existing code, e.g. in a user's visual plugin, that looks up depth cameras by name.

chapulina:
Closing this since it can't be merged in its current form. Feel free to open another PR with a proposal that's backwards-compatible.

PR #2737, opened 2020-05-15
[setup.sh] preserve GAZEBO_MASTER_URI if set

moriarty:
If GAZEBO_MASTER_URI is set use that otherwise use the default.

Most of the environment variables set in setup.sh already preserve or append to values which were already set.

Example use case:
If you want to run `gzserver` and `gzclient` in two different containers. You would first start the `gzserver` and then pass the GAZEBO_MASTER_URI to the `docker run -e` for the `gzclient` which could now source `setup.sh` without needing it's own custom entrypoint.

The only environment variable in `setup.sh` which does not preserve or append is `GAZEBO_MODEL_DATABASE_URI` - I imagine it may be worth doing the same change to that variable too.


moriarty:
wow that's a lot of :heavy_multiplication_x: checks failing

j-rivero:

scpeters (approved):

j-rivero:
all builds are UNSTABLE (expected at this moment) and test failures are not relevant to this PR as far as I can see.

I thought the change would bring some bash particular syntax to the file but seems like variable default values are in [the posix standard](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html), good.

j-rivero (approved):

j-rivero:
Thanks @moriarty ! I final request, could you please update the Changelog.md file with an entry about this change?

j-rivero:
Thanks Alex!

PR #2745, opened 2020-05-29
Rename io.h to avoid problems with Windows (gazebo11)

j-rivero:
The Windows failure in Gazebo11 seems to happen because we are introducing an internal io.h file inside the compilation and Windows is unable to get its own header with the same name. I don't think we are exporting the header or using it in any other place than the internal ode fork so I'm changing its name here.


j-rivero:
@osrf-jenkins run tests

j-rivero:

j-rivero:

j-rivero:
I've fixed some problems with Windows setup, running now [Build Status](https://build.osrfoundation.org/job/gazebo-ci-pr_any-windows7-amd64/2773/)

Rest of testing failures in other platforms seems unrelated to me.

nuclearsandwich (approved):

Issue #2735, opened 2020-05-13
Invalid Libs flags in pkg-config files from OSRF Gazebo packages for Focal


doudou:
The Gazebo packages for focal (as distributed by OSRF) have invalid `Libs` flags (both gazebo9 and gazebo11 packages) in `gazebo.pc`

They list the boost CMake targets as `-l` flags instead of either the proper `-l` flags or the full path to the libraries:

~~~
lBoost::thread -lBoost::system -lBoost::filesystem -lBoost::program_options -lBoost::regex -lBoost::iostreams -lBoost::date_time
~~~

Versions:
- libgazebo9-dev: 9.13.0-1~focal
- libgazebo11-dev: 11.0.0-2~focal

PR #2751, opened 2020-06-08
Fix pkg-config boost entries for Ubuntu Focal (gazebo9)

j-rivero:
Partial fix for https://github.com/osrf/gazebo/issues/2735 . Need another PR for Gazebo11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -386,6 +386,8 @@ else (build_errors)
     endif()
     # Remove the prefix lib (not always present, like in pthread)
     string (REPLACE "lib" "" bname "${bname}")
+    # Some boost versions add the Boost:: component toe cmake, need to change that to boost_ prefix
+    string (REPLACE "Boost::" "boost_" bname "${bname}")
     set (Boost_PKGCONFIG_LIBS "${Boost_PKGCONFIG_LIBS} ${bname}")
   endforeach(b)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ set (GAZEBO_MINOR_VERSION 13)
 # The patch version may have been bumped for prerelease purposes; be sure to
 # check gazebo-release/ubuntu/debian/changelog@default to determine what the
 # next patch version should be for a regular release.
-set (GAZEBO_PATCH_VERSION 1)
+set (GAZEBO_PATCH_VERSION 2)
 
 set (GAZEBO_VERSION ${GAZEBO_MAJOR_VERSION}.${GAZEBO_MINOR_VERSION})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -386,6 +386,8 @@ else (build_errors)
     endif()
     # Remove the prefix lib (not always present, like in pthread)
     string (REPLACE "lib" "" bname "${bname}")
+    # Some boost versions add the Boost:: component toe cmake, need to change that to boost_ prefix
```
scpeters: `toe` -> `to`

scpeters (approved):
I've noticed this issue for a while on homebrew since it has had a newer boost version for a while. Unfortunately, this doesn't fix the issue on homebrew because those libraries have an extra `-mt` (for multithreaded I believe).

~~~
$ ls /usr/local/opt/boost/lib | grep thread
libboost_thread-mt.a
libboost_thread-mt.dylib
~~~

The homebrew CI build is gone now, but I saw the following in an email about it:

~~~
399: ld: library not found for -lboost_thread
~~~

In my previous debugging, I tried to use a generator expression to extract the boost library names from cmake targets. For example, the following patch is able to write the absolute path of the library for each target:

~~~
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8659ff1707..e90e868a23 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -383,6 +383,7 @@ else (build_errors)
   string (REPLACE ";" " -L" Boost_PKGCONFIG_LIBS "${Boost_PKGCONFIG_LIBS}")
 
   foreach (b ${Boost_LIBRARIES})
+    file(GENERATE OUTPUT generated_${b}.txt CONTENT $<TARGET_LINKER_FILE:${b}>\n)
     get_filename_component(bname ${b} NAME_WE)
     # Prefix -l if not already prefixed
     string(FIND ${bname} "-l" lprefix)
~~~

~~~
$ cat generated_Boost\:\:thread.txt 
/usr/local/lib/libboost_thread-mt.dylib
~~~

cmake has a different generator expression that seems like it would be perfect for this case: `TARGET_LINKER_FILE_BASE_NAME`, but that been reported as a known issue for these boost targets (https://github.com/boostorg/boost_install/issues/41, and [cmake issue 20611](https://gitlab.kitware.com/cmake/cmake/-/issues/20611))

so I'll approve this because it seems to work for ubuntu 20.04, but it doesn't fix homebrew

CMakeLists.txt
scpeters: fixed in a7f0b0fae10ec10cae3a2205e9471b64ea0fc1f2

scpeters:
I've updated the changelog and also bumped the version to 9.13.2 in anticipation of an upcoming release

scpeters:
I filed #2790 to capture the issues with homebrew that I discussed in that comment

PR #2764, opened 2020-06-25
LensFlare: initialize OGRE compositors during plugin initialization

scpeters:
There are several `SensorPlugin`s that create Ogre compositors (such as [LensFlareSensorPlugin](https://github.com/osrf/gazebo/blob/gazebo11/plugins/LensFlareSensorPlugin.cc)). It was reported by @mogumbo that the order in which compositors are applied affects the rendered output, but the `LensFlareSensorPlugin` always seems to be applied last, regardless of the order in which plugins are specified in an SDFormat file. @iche033 suggested moving the code for initializing a `gazebo::rendering::LensFlare` compositor from `LensFlare::Update` to `Lensflare::SetCamera`, so that the compositor would be added during plugin initialization.

The initialization code expects a pointer to a `DirectionalLight`, which is not available until the first render update. It seems to work to initialize with a `nullptr` for the `DirectionalLight` since the light ptr is being set every update. I also added logic to exit `notifyMaterialRender` early if the light pointer is `nullptr`.

### DIFF
--- a/gazebo/rendering/LensFlare.cc
+++ b/gazebo/rendering/LensFlare.cc
@@ -76,6 +76,12 @@ namespace gazebo
       public: virtual void notifyMaterialRender(unsigned int _passId,
                                                 Ogre::MaterialPtr &_mat)
       {
+        if (!this->light)
+        {
+          // return if this->light is not set, we may still be initializing
+          return;
+        }
+
         GZ_ASSERT(!_mat.isNull(), "Null OGRE material");
         // These calls are setting parameters that are declared in two places:
         // 1. media/materials/scripts/gazebo.material, in
@@ -96,12 +102,6 @@ namespace gazebo
         params->setNamedConstant("viewport",
             Ogre::Vector3(static_cast<double>(this->camera->ViewportWidth()),
             static_cast<double>(this->camera->ViewportHeight()), 1.0));
-
-        if (!this->light)
-        {
-          // return if this->light is not set, we may still be initializing
-          return;
-        }
 
         // use light's world position for lens flare position
         if (this->light->Type() == "directional")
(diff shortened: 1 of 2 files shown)


gazebo/rendering/LensFlare.cc
```diff
@@ -97,6 +97,12 @@ namespace gazebo
             Ogre::Vector3(static_cast<double>(this->camera->ViewportWidth()),
             static_cast<double>(this->camera->ViewportHeight()), 1.0));
 
+        if (!this->light)
```
iche033: I think you should be able move this check to the top of the function as we don't need to set the `params` anymore

gazebo/rendering/LensFlare.cc
```diff
@@ -440,31 +472,6 @@ void LensFlare::Update()
...
-        this->dataPtr->camera->OgreViewport(), "CameraLensFlare/Default");
-    this->dataPtr->lensFlareInstance->getTechnique()->getOutputTargetPass()->
-        getPass(0)->setMaterial(lensFlareMaterial);
-
-    this->dataPtr->lensFlareInstance->setEnabled(true);
-    this->dataPtr->lensFlareInstance->addListener(
-        this->dataPtr->lensFlareCompositorListener.get());
-  }
-  else
   {
```
iche033: remove brackets?

iche033 (approved):
looks good to me. Just very minor comments

gazebo/rendering/LensFlare.cc
scpeters: 74c0894f4a8fa971b369b18914e1748977028f3d

gazebo/rendering/LensFlare.cc
scpeters: 74c0894f4a8fa971b369b18914e1748977028f3d

PR #2773, opened 2020-07-01
Fix Actor collision if loop / auto_start false

vvasco:
As reported [here](https://answers.gazebosim.org/question/25409/actor-collision-when-loop-is-false/), when using the Actor Collision Plugin with [this example](https://github.com/osrf/gazebo/tree/gazebo11/examples/plugins/actor_collisions), if `loop` (or `auto_start`) is set to `false`, the boxes around the actor links representing the collisions start falling around. 
The problem is that in such cases, the pose of the actor is not set.
This PR proposes a fix, by setting the pose of the actor based on the last animated frame when it stops.
This is the result:

actor-coll


chapulina:
Thank you for the pull request! The code looks reasonable. I'll give it a try in the coming weeks.

chapulina (approved):
Thanks again for the PR. Works for me!

I'm just going to trigger CI one more time to make sure all tests are ok. If you're motivated, it would be good to add a test to [Actor_TEST](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/physics/Actor_TEST.cc) exercising the fixed functionality.

chapulina:

vvasco:
Sure, I will add the test.

vvasco:
I added a test to `Actor_TEST`, which loads the [actor collision scenario](https://github.com/vvasco/gazebo/blob/fix/actor-collision-plugin/examples/plugins/actor_collisions/actor_collisions_noloop.world) with `loop` set to `false` and :
- checks that collisions are enabled;
- checks that collisions remain enabled when the actor is inactive;
- checks that links poses do not change when the actor is inactive.

gazebo/physics/Actor_TEST.cc
```diff
@@ -102,6 +102,62 @@ TEST_F(ActorTest, TrajectoryFromSDF)
...
+                                        "LeftFoot", "RHipJoint", "RightUpLeg",
+                                         "RightLeg", "RightFoot", "LowerBack",
+                                         "Spine", "Neck", "Neck1",
+                                         "LeftShoulder", "RightShoulder",
+                                         "LeftArm", "LeftForeArm",
+                                         "RightArm", "RightForeArm",
+                                         "LeftFingerBase", "RightFingerBase"};
+  for (const auto &name : linkNames)
+  {
+    auto link = actor->GetLink(name);
```
chapulina: Just in case something changes in the actor model, it would be good to `ASSERT_NE(nullptr, link);` before using the link pointer below. Otherwise the test will crash.

gazebo/physics/Actor_TEST.cc
```diff
@@ -102,6 +102,62 @@ TEST_F(ActorTest, TrajectoryFromSDF)
...
+    ASSERT_GT(link->GetCollisions().size(), 0u);
```
chapulina: This can be an `EXPECT_GT`, because it's ok for the test to continue even if this check fails.

gazebo/physics/Actor_TEST.cc
```diff
@@ -102,6 +102,62 @@ TEST_F(ActorTest, TrajectoryFromSDF)
   EXPECT_LT(fabs(actor->ScriptTime() - world->SimTime().Double()), 1.0 / 30);
 }
 
+//////////////////////////////////////////////////
+TEST_F(ActorTest, ActorCollision)
+{
+  // Load a world with an actor
+  this->Load("examples/plugins/actor_collisions/actor_collisions_noloop.world",
```
chapulina: The problem of loading this world in the test is that it assumes that the plugin is compiled and in the path. The test is usually ran in environments where this isn't true. For me, it printed this error and the test failed:

```
[Err] [Plugin.hh:178] Failed to load plugin libActorCollisionsPlugin.so: libActorCollisionsPlugin.so: cannot open shared object file: No such file or directory
```

I think we have 2 options:

1. Compile the plugin during the test. This can become tough, you'd need to do something [like this](https://github.com/osrf/gazebo/blob/gazebo11/test/examples/examples_build.cc), then add it to `GAZEBO_PLUGIN_PATH`, then be sure to cleanup at the end.
2. Change this test to just check the changes to `Actor.cc`. For example:
    * create a simple test world in [test/worlds](https://github.com/osrf/gazebo/tree/gazebo11/test/worlds) with false `loop` / `auto_start`
    * step until the animation ends
    * change the actor's pose from the test (to simulate what gravity would do in case the plugin had added collisions)
    * step a bit more, then check that the actor went back to the pose at the end of the loop

chapulina (commented):
Thank you for adding the test! Some changes are needed for it to work on all systems, see below.

gazebo/physics/Actor_TEST.cc
vvasco: Ok thank you, I would go for option 2 then.
So the test world will not include anymore the collision plugin, therefore I can remove the part of the test that checks if collisions are enabled.

vvasco:
I updated the new version of the test.

chapulina (approved):
Thank you for iterating, this is good to go!

PR #2776, opened 2020-07-03
[Gazebo 11] Added ignition common profiler

ahcorde:
I added profile points in ODEPhysics, sensors and some plugins

Here some snapshots:

Selección_044
Selección_042
Selección_041

there is a world called `profile.world` where you can reproduce these results.

I had to change this line in the CMakeLists.txt to avoid create a Profile Singleton in each library. I can provide more details about this if necessary.

from
```cmake
  filter_valid_compiler_flags(-fvisibility=hidden -fvisibility-inlines-hidden)
```
to
```cmake
  filter_valid_compiler_flags(-fvisibility-inlines-hidden)
```

is this change creating a conflict with something specific?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,7 +260,7 @@ filter_valid_compiler_flags(${WARN_LEVEL}
 # Check and add visibility hidden by default. Only in UNIX
 # Windows and MacosX does not handled properly the hidden compilation
 if (UNIX AND NOT APPLE)
-  filter_valid_compiler_flags(-fvisibility=hidden -fvisibility-inlines-hidden)
+  filter_valid_compiler_flags(-fvisibility-inlines-hidden)
 endif()
 
 if (MSVC)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,14 @@ if(NOT DEFINED BUILD_TESTING)
     set(BUILD_TESTING OFF)
 endif()
 
+option(ENABLE_PROFILER "Enable Ignition Profiler" FALSE)
+
+if(ENABLE_PROFILER)
+  add_definitions("-DIGN_PROFILER_ENABLE=1")
+else()
+  add_definitions("-DIGN_PROFILER_ENABLE=0")
+endif()
+
 #============================================================================
 # We turn off extensions because (1) we do not ever want to use non-standard
 # compiler extensions, and (2) this variable is on by default, causing cmake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,7 +268,11 @@ filter_valid_compiler_flags(${WARN_LEVEL}
 # Check and add visibility hidden by default. Only in UNIX
 # Windows and MacosX does not handled properly the hidden compilation
 if (UNIX AND NOT APPLE)
-  filter_valid_compiler_flags(-fvisibility-inlines-hidden)
+  if (ENABLE_PROFILER)
+    filter_valid_compiler_flags(-fvisibility-inlines-hidden)
+  else()
+    filter_valid_compiler_flags(-fvisibility=hidden -fvisibility-inlines-hidden)
+  endif()
 endif()
 
 if (MSVC)
--- a/gazebo/Server.cc
+++ b/gazebo/Server.cc
@@ -33,6 +33,7 @@
 #include <sdf/sdf.hh>
 
 #include <ignition/math/Rand.hh>
+#include "ignition/common/Profiler.hh"
 
 #include "gazebo/gazebo.hh"
 #include "gazebo/transport/transport.hh"
@@ -582,16 +583,25 @@ void Server::Run()
 
   this->dataPtr->initialized = true;
 
+  IGN_PROFILE_THREAD_NAME("gzserver");
   // Stay on this loop until Gazebo needs to be shut down
   // The server and sensor manager outlive worlds
   while (!this->dataPtr->stop)
   {
+    IGN_PROFILE("gzserver_loop");
+    IGN_PROFILE_BEGIN("ProcessControlMsgs");
     this->ProcessControlMsgs();
+    IGN_PROFILE_END();
 
-    if (physics::worlds_running())
+    if (physics::worlds_running()) {
+      IGN_PROFILE_BEGIN("sensors::run_once");
       sensors::run_once();
-    else if (sensors::running())
+      IGN_PROFILE_END();
+    } else if (sensors::running()) {
+      IGN_PROFILE_BEGIN("sensors::stop");
       sensors::stop();
+      IGN_PROFILE_END();
+    }
 
     common::Time::MSleep(1);
   }
--- a/gazebo/Server.cc
+++ b/gazebo/Server.cc
@@ -588,17 +588,20 @@ void Server::Run()
   // The server and sensor manager outlive worlds
   while (!this->dataPtr->stop)
   {
-    IGN_PROFILE("gzserver_loop");
+    IGN_PROFILE("Server::Run");
     IGN_PROFILE_BEGIN("ProcessControlMsgs");
     this->ProcessControlMsgs();
     IGN_PROFILE_END();
 
-    if (physics::worlds_running()) {
-      IGN_PROFILE_BEGIN("sensors::run_once");
+    if (physics::worlds_running())
+    {
+      IGN_PROFILE_BEGIN("run_once");
       sensors::run_once();
       IGN_PROFILE_END();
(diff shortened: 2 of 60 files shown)


CMakeLists.txt
```diff
@@ -260,7 +268,7 @@ filter_valid_compiler_flags(${WARN_LEVEL}
 # Check and add visibility hidden by default. Only in UNIX
 # Windows and MacosX does not handled properly the hidden compilation
 if (UNIX AND NOT APPLE)
-  filter_valid_compiler_flags(-fvisibility=hidden -fvisibility-inlines-hidden)
+  filter_valid_compiler_flags(-fvisibility-inlines-hidden)
```
iche033: maybe @j-rivero can comment on whether this change would be a problem or not?

gazebo/Server.cc
```diff
@@ -582,16 +583,25 @@ void Server::Run()
...
   // The server and sensor manager outlive worlds
   while (!this->dataPtr->stop)
   {
+    IGN_PROFILE("gzserver_loop");
+    IGN_PROFILE_BEGIN("ProcessControlMsgs");
     this->ProcessControlMsgs();
+    IGN_PROFILE_END();
 
-    if (physics::worlds_running())
+    if (physics::worlds_running()) {
```
iche033: bracket on new line. 

on a related note, I get a few code check errors from running `sh tools/code_check.sh`

gazebo/sensors/Sensor.cc
```diff
@@ -225,6 +227,9 @@ bool Sensor::NeedsUpdate()
 //////////////////////////////////////////////////
 void Sensor::Update(const bool _force)
 {
+  // //IGN_PROFILE_THREAD_NAME("Sensor");
+  //IGN_PROFILE("Sensor");
+  //IGN_PROFILE_BEGIN("Sensor::Update");
```
iche033: remove?

gazebo/sensors/SensorManager.cc
```diff
@@ -546,8 +548,13 @@ void SensorManager::SensorContainer::RunLoop()
   else
     sleepTime.Set(0, 1e6);
 
+  IGN_PROFILE_THREAD_NAME("SensorManager");
+
   while (!this->stop)
   {
+    IGN_PROFILE("SensorManagerLoop");
+    // IGN_PROFILE_BEGIN("SensorManagerLoop");
```
iche033: remove?

gazebo/sensors/SensorManager.cc
```diff
@@ -595,10 +604,14 @@ void SensorManager::SensorContainer::RunLoop()
...
 
     // This if statement helps prevent deadlock on osx during teardown.
+    IGN_PROFILE_BEGIN("Sleeping");
     if (!this->stop)
     {
       this->runCondition.wait(timingLock);
     }
+    IGN_PROFILE_END();
+
+    // IGN_PROFILE_END();
```
iche033: remove?

gazebo/physics/ode/ODEPhysics.cc
```diff
@@ -352,13 +353,16 @@ void ODEPhysics::Init()
...
+  IGN_PROFILE_THREAD_NAME("ODEPhysics");
   dAllocateODEDataForThread(dAllocateMaskAll);
 }
 
 //////////////////////////////////////////////////
 void ODEPhysics::UpdateCollision()
 {
   DIAG_TIMER_START("ODEPhysics::UpdateCollision");
+  IGN_PROFILE("ODEPhysics:UpdateCollision");
+  IGN_PROFILE_BEGIN("dSpaceCollide");
```
iche033, line 365: In other places of this PR, I see the the profile name prefixed by the class name. For example `AltimeterSensor::publish`. But that's also strange since it makes me think that `publish` is a function belonging `AltimeterSensor`, which is not the case. Looking at ign-gazebo, I see a convention like `class_name::function_name profile_name`, e.g. [here](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/rendering/RenderUtil.cc#L329). Maybe we can follow the same convention? Alternatively, just do what you have here now, i.e. `dSpaceCollide` without the class name prefix name, and update other places in the PR to keep them consistent.

iche033 (commented):
works for me. I have some minor comments about style and profile names.

gazebo/physics/ode/ODEPhysics.cc
ahcorde: Fixed and added your feedback https://github.com/osrf/gazebo/pull/2776/commits/f8d068fb7400597ff0d21425063107bee714ade0

iche033:
latest changes look good to me. Just waiting to see if @j-rivero has any thoughts on whether or not it's ok to remove the `-fvisibility=hidden` compiler flag

gazebo/Server.cc
```diff
@@ -582,16 +583,28 @@ void Server::Run()
...
+      IGN_PROFILE_END();
+    }
     else if (sensors::running())
+    {
+      IGN_PROFILE_BEGIN("stop");
       sensors::stop();
+      IGN_PROFILE_END();
+    }
 
     common::Time::MSleep(1);
```
scpeters: hmm...having a 1 ms sleep in the server runloop seems non-ideal

chapulina: I remember seeing this before and wondering why it's there :monocle_face:

iche033: I saw this while working on lockstepping. I tried removing the line but it caused some sensor tests to fail so I had to put it back in:
https://github.com/osrf/gazebo/pull/2746/files#diff-2e1da9464c22259ef9d9919412c42d2fR617

scpeters: I wonder if we can we change it to `NSleep(1)`?

this should happen in a separate PR; I'll try creating a PR with this change

CMakeLists.txt
ahcorde: friendly ping @j-rivero

j-rivero: uh, the change is important. Symbols are exposed by default in the ABI unless the `fvisibility=hidden` flag is provided. This change will increase the ABI size (with an impact on the loader performance) and expose many new symbols to the binary interface. Can we make it conditional somehow to keep the current ABI scheme?

ahcorde: Added https://github.com/osrf/gazebo/pull/2776/commits/a9230be4b6fc885148bb001ed30939f8a8d26c26

iche033:
latest changes look good. There are a few new code check errors found by running `sh tools/code_check.sh`

ahcorde:
@iche033 Any trick to show only the new ones?

iche033:
oh I don't know a trick to show just the new errors. I just manually ran code check and checked the errors against the code. Here're the ones I see:

```
./plugins/LiftDragPlugin.cc:404:  Blank line at the end of a code block.  Is this needed?  [whitespace/blank_line] [3]
./plugins/LinkPlot3DPlugin.cc:207:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/sensors/WideAngleCameraSensor.cc:185:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/sensors/SensorManager.cc:19:  Found C++ system header after other header. Should be: c system, c++ system, other, SensorManager.h .  [build/include_order] [4]
./gazebo/sensors/SensorManager.cc:20:  Found C++ system header after other header. Should be: c system, c++ system, other, SensorManager.h .  [build/include_order] [4]
./gazebo/physics/World.cc:40:  Found C++ system header after other header. Should be: c system, c++ system, other, World.h .  [build/include_order] [4]
./gazebo/common/Event.hh:32:  Found C++ system header after other header. Should be: c system, c++ system, other, Event.h .  [build/include_order] [4]
./gazebo/common/Event.hh:280:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/common/Event.hh:300:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/common/Event.hh:321:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/common/Event.hh:343:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/common/Event.hh:367:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/common/Event.hh:393:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/common/Event.hh:420:  { should never be at the end of the previous line  [whitespace/braces] [4]
./gazebo/common/Event.hh:448:  { should never be at the end of the previous line  [whitespace/braces] [4]
```

iche033 (approved):

PR #2778, opened 2020-07-05
Fix assumptions that CMAKE_INSTALL_*DIR paths are relative.

lopsided98:
When an absolute path is desired, the CMAKE_INSTALL_FULL_*DIR variables should be used instead of concatenating with CMAKE_INSTALL_PREFIX. Special handling is needed for pkg-config files to ensure they use interpolations when possible.

See https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path for more information.


j-rivero:
I think that the change proposed is correct, it uses the right variables from [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html). We need to verify that values in pkgconfig and cmake files looks correct when they are generated.

mjcarroll:
@osrf-jenkins run tests

mjcarroll (approved):
Changes LGTM, thanks for the contribution.

I'm going to let this run through CI again before merging, it looks like it may have been infrastructure issues before.

mjcarroll:
@osrf-jenkins run tests

mjcarroll:
@osrf-jenkins run tests

mjcarroll:
Really struggling to get CI to turn over here, I checked build.osrfoundation.org before kicking this one off, so hopefully we'll be in better shape this time.

PR #2784, opened 2020-07-10
Fix sensor update rate throttling when new sensors are spawned

iche033:
There is an issue with sensor manager's update loop throttling. When a world is loaded, the sensor manager figures out the max update rate from all the sensors in the world. However, when a new sensor is spawned after the world is loaded, this max update rate is never updated. This causes a problem for example when later spawning a sensor with an update rate higher than all the existing sensors in the world, this new sensor would never hit its target update rate.

This PR adds a flag (global variable, `g_sensorsDirty`, to avoid breaking ABI) to detect when sensors are added or removed, and recalcuates the sensor manager's max update rate when the flag is dirty.

The bug only affects non-rendering sensors, e.g. imu, ray sensor, etc

To manually test this, you can:
1. download the `hokuyo` model from gazebo model database, e.g. by inserting it from gazebo's Insert tab on the left
1. create a copy with higher update rate, e.g. copy the model `~/.gazebo/models/hokuyo` to `~/.gazebo/models/hokuyo_new` and change the update rate in `model.sdf` and also remember to give it a new name in `model.config`
1. launch gazebo and insert the original hokuyo model -> check update rate, e.g. Ctrl + T and select the corresponding laser scan topic under `LaserStamped` msg type
1. insert the new hokuyo model with higher update rate -> check update rate using the same way as before. Verify that the new model is able to reach the target update rate with the changes in this PR.

### DIFF
--- a/gazebo/rendering/Visual.cc
+++ b/gazebo/rendering/Visual.cc
@@ -818,10 +818,9 @@ void Visual::SetScale(const ignition::math::Vector3d &_scale)
     this->dataPtr->sceneNode->setScale(
         Conversions::Convert(this->dataPtr->scale));
   }
-  else
-  {
-    gzerr << Name() << ": Size of the collision contains one or several zeros."
-      << " Collisions may not visualize properly." << std::endl;
+  else {
+    gzerr << Name() << ": Size of the collision contains one or several zeros." <<
+      " Collisions may not visualize properly." << std::endl;
   }
   // Scale selection object in case we have one attached. Other children were
   // scaled from UpdateGeomSize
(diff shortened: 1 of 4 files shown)


gazebo/rendering/Visual.cc
```diff
@@ -817,9 +817,11 @@ void Visual::SetScale(const ignition::math::Vector3d &_scale)
   {
     this->dataPtr->sceneNode->setScale(
         Conversions::Convert(this->dataPtr->scale));
-  } else {
-    gzerr << Name() << ": Size of the collision contains one or several zeros." <<
-      " Collisions may not visualize properly." << std::endl;
+  }
```
ahcorde: This change is not related with the PR title. I introduced this, I can open a PR to fix it.

ahcorde (approved):
LGTM, just remove the changes in `gazebo/rendering/Visual.cc`.  i will open a PR to fix it separately

gazebo/rendering/Visual.cc
iche033: done, reverted these changes

ahcorde (approved):
Fixed brackets https://github.com/osrf/gazebo/pull/2784

Issue #2611, opened 2019-03-27
[Gazebo9][Windows] ModelDatabase::GetModelPath implementation is incomplete on Windows build.


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/2611)) by Sean Yen (Bitbucket: [seanyen-msft](https://bitbucket.org/%7B9f25a1e7-e37c-4707-a084-64e8be6f3785%7D/)).**

----------------------------------------
https://github.com/osrf/gazebo/blob/ab82c04da4afe36cb7e6b9842a1d7b4cdfa35570/gazebo/common/ModelDatabase.cc#L506

I noticed that the GetModelPath requires libtar and maybe because of the lack of libtar port on Windows, some code path is disabled on Windows build, which breaks the scenario to use online models.

I'd like to use this bug to track how we should address this for Windows build.

traversaro:
While using `libarchive`  is probably a cleaner version, @seanyen solution of using the `tar` command on Windows added in https://github.com/osrf/gazebo/pull/2789/files#diff-8f453ff422db23b46311978f86b355af2060a7833e4201045d5112fd2d502990 seems to be working fine, so it probably make sense to propagate it to Gazebo 11 as well (were it is missing, see https://github.com/osrf/gazebo/blob/gazebo11_11.5.0/gazebo/common/ModelDatabase.cc#L506 that is missing the `#else` block w.r.t. to https://github.com/osrf/gazebo/blob/gazebo9_9.17.0/gazebo/common/ModelDatabase.cc#L509 . 

TL;DR: To solve this, the simplest solution is to forward propagate https://github.com/osrf/gazebo/pull/2789 to the `gazebo11` branch.

PR #2789, opened 2020-07-14
[Gazebo9] More enhancement for Windows build

seanyen:
* Installing the DLLs (shared libraries) into the `bin` folder by convention.
* Normalize the material paths into native paths so OGRE can discover them later. (Mixing the forward/backward slashes really confuses OGRE when looking up the resources.)
* Use the built-in `tar.exe` in Windows 10 (yes, there is a [`tar.exe`](https://techcommunity.microsoft.com/t5/containers/tar-and-curl-come-to-windows/ba-p/382409) with Windows) to unpack the downloaded asset. This could be a way to resolve this gap: #2611

### DIFF
--- a/cmake/GazeboUtils.cmake
+++ b/cmake/GazeboUtils.cmake
@@ -131,7 +131,11 @@ endmacro()
 #################################################
 macro (gz_install_library _name)
   set_target_properties(${_name} PROPERTIES SOVERSION ${GAZEBO_MAJOR_VERSION} VERSION ${GAZEBO_VERSION_FULL})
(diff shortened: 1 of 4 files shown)


j-rivero:
Thanks Sean! changes looks good to me. Let me run another round of CI before merging. @osrf-jenkins run tests

j-rivero:
The abi-checker killed the memory in CI, I'm not worried about. The Mac build is waiting #2809. Windows is happy. Let's wait a bit more for the linux ones.

seanyen:
Thanks! And hope this can be forward-port to Gazebo11 if they also make sense there. :)

j-rivero (approved):

chapulina (approved):

PR #2788, opened 2020-07-13
[Gazebo 9] Added test to check collisions equal to zero

ahcorde:
I missed to add tests in this PR https://github.com/osrf/gazebo/pull/2768. 

I added a world with a box, sphere and cylinder with collisions equal to zero. The test will try to load the world, no segmentation fault should happens

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #2770, opened 2020-06-29
[Forwardport Gazebo11] Fixed crash when collision size is zero

ahcorde:
Orignal PR #2768

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #2795, opened 2020-07-23
[Forwardport Gazebo 11] Added test to check collisions equal to zero (#2788)

ahcorde:
Forwardport https://github.com/osrf/gazebo/pull/2788


mabelzhang (approved):
Worked for me.
I tested by reverting #2770, saw that the test fails, then re-applying that PR, and saw that the test passes.
Looks like the buildfarm webpage is unresponsive again, so I can't see CI results. Since this PR is a forward port and only adding a new test, I think it's reasonable to say it shouldn't break other tests.

PR #2796, opened 2020-07-23
Merge forward gazebo9 -> gazebo10

j-rivero:


ahcorde:
I compiled this branch locally and I'm getting this error:

```
/usr/include/dart/dynamics/detail/CompositeNode.hpp:52:15: error: ‘make_unique’ is not a member of ‘std’
   52 |   return std::make_unique<State>(common::Composite::getCompositeState());
      |               ^~~~~~~~~~~
/usr/include/dart/dynamics/detail/CompositeNode.hpp:52:15: note: ‘std::make_unique’ is only available from C++14 onwards
/usr/include/dart/dynamics/detail/CompositeNode.hpp:52:32: error: expected primary-expression before ‘>’ token
   52 |   return std::make_unique<State>(common::Composite::getCompositeState());
      |                                ^
/usr/include/dart/dynamics/detail/CompositeNode.hpp: In member function ‘std::unique_ptr<dart::dynamics::Node::Properties, std::default_delete<dart::dynamics::Node::Properties> > dart::dynamics::CompositePropertiesNode<Base>::getNodeProperties() const’:
/usr/include/dart/dynamics/detail/CompositeNode.hpp:77:15: error: ‘make_unique’ is not a member of ‘std’
   77 |   return std::make_unique<Properties>(
      |               ^~~~~~~~~~~
/usr/include/dart/dynamics/detail/CompositeNode.hpp:77:15: note: ‘std::make_unique’ is only available from C++14 onwards
/usr/include/dart/dynamics/detail/CompositeNode.hpp:77:37: error: expected primary-expression before ‘>’ token
   77 |   return std::make_unique<Properties>(

```

is there any specific version of DART to compile with Gazebo 10 ? 

This line is forcing C++11 when it should be C++14. 

https://github.com/osrf/gazebo/blob/84489bb0211007e3e45e472e7f7b90293187cf76/CMakeLists.txt#L268

But this line it's also in Gazebo 11 and it's compiling, 

https://github.com/osrf/gazebo/blob/d33e6429e3aa63181c3cef6d292370e6dc3f4320/CMakeLists.txt#L274

I don't know waht is happening here. any ideas @j-rivero ?

chapulina:
> This line is forcing C++11 when it should be C++14.

We shouldn't change the minimum C++ version for a Gazebo version that's already released. We should use a DART version that's C++11-compatible instead.

chapulina:

chapulina (approved):
The forward-port looks good to me. Ubuntu and Windows CI were passing, and Homebrew is failing due to infrastructure problems.

I just triggered new tests to double-check that no new test failures are introduced, because the previous tests weren't available anymore.

scpeters:
there is a complaint in the ABI checker about a change in layout of the v-table of DepthCameraPlugin:

* https://build.osrfoundation.org/job/gazebo-abichecker-any_to_any-ubuntu_auto-amd64/431/API_5fABI_20report/

PR #2812, opened 2020-08-04
Fix race condition on Publisher shutdown

iche033:
The race condition problem occurs when the `Publisher::OnPublishComplete` callback is invoked from a different thread after the `Publisher` object is destroyed, causing a crash. There's been a few attempts to fixing this race condition, see this [commit](https://github.com/osrf/gazebo/commit/e87d6cd82cede31b62025b367571bea2424ca602) and this [PR](https://bitbucket.org/osrf-migrated/gazebo/pull-requests/3103/fix-race-conditions-in-master/diff), but the issue persists and it is just harder to reproduce. When successfully reproduced, I get a assertion error in this [line](https://github.com/osrf/gazebo/blob/gazebo9/gazebo/transport/Publisher.cc#L242) when locking the mutex.

The fix is to bind the callback function along with a shared pointer instead of raw pointer `this` so that the object is kept alive for a little longer until after the callback is complete. In the case of Publisher, this should be safe to do as the `OnPublishComplete` function is mainly just doing some bookkeeping work. Note that I added a `PublisherPrivate` class and stored instances of this class in a static variable to avoid breaking ABI.

To reproduce this issue:

1. Save this model as `box.sdf` file:

```
<?xml version="1.0" ?>
<sdf version="1.6">
  <model name="box">
    <pose>0 0 0.5 0 0 0</pose>
    <link name="link">
      <collision name="collision">
        <geometry>
          <box>
            <size>1 1 1</size>
          </box>
        </geometry>
      </collision>
      <visual name="visual">
        <geometry>
          <box>
            <size>1 1 1</size>
          </box>
        </geometry>
      </visual>
    </link>
  </model>
</sdf>
```

2. launch gazebo using roslaunch: 

```
roslaunch gazebo_ros empty_world.launch verbose:=true
```

3. Run the following script that spawns and deletes the box model 500 times:

```
#!/bin/bash
for i in {0..500}; do
    echo "$i"
    rosparam set pizza --textfile box.sdf
    rosrun gazebo_ros spawn_model -param pizza -sdf -model box -x 0 -y 0.5 -z 0
    sleep 0.5
    rosservice call gazebo/delete_model '{model_name: box}'
done
```

4. Without changes in this PR, the crash occurs on spawn after a couple hundred iterations for me

### DIFF
--- a/gazebo/transport/Publisher.cc
+++ b/gazebo/transport/Publisher.cc
@@ -31,6 +31,48 @@ using namespace transport;
 
 uint32_t Publisher::idCounter = 0;
 
+namespace gazebo
+{
+/// \brief Private class for Publisher
+class PublisherPrivate
+{
+  /// \brief Callback when a publish is completed
+  /// \param[in] _id ID associated with the publication.
+  public: void OnPublishComplete(uint32_t _pubId);
+
+  /// \brief Current publication ids.
+  public: std::map<uint32_t, int> pubIds;
+
+  /// \brief Mutex to protect pubIds
+  public: std::mutex mutex;
+};
+}
+
+/// \brief Mutex to protect publisherMap
+static std::mutex pubMapMutex;
+
+/// \brief A map of Publisher id and its shared pointer
+static std::map<uint32_t, std::shared_ptr<PublisherPrivate>> publisherMap;
+
+//////////////////////////////////////////////////
+void PublisherPrivate::OnPublishComplete(uint32_t _id)
+{
+  try
+  {
+    // This is the deeply unsatisfying way of dealing with a race
+    // condition where the publisher is destroyed before all
+    // OnPublishComplete callbacks are fired.
+    std::lock_guard<std::mutex> lock(this->mutex);
+    std::map<uint32_t, int>::iterator iter = this->pubIds.find(_id);
+    if (iter != this->pubIds.end() && (--iter->second) <= 0)
+      this->pubIds.erase(iter);
+  }
+  catch(...)
+  {
+    return;
+  }
+}
+
 //////////////////////////////////////////////////
 Publisher::Publisher(const std::string &_topic, const std::string &_msgType,
                      unsigned int _limit, double _hzRate)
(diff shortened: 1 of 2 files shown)


gazebo/transport/Publisher.cc
```diff
@@ -188,12 +250,14 @@ void Publisher::SendMessage()
     {
       // Send the latest message.
       int result = this->publication->Publish(*iter,
-          boost::bind(&Publisher::OnPublishComplete, this, _1), *pubIter);
+          boost::bind(&PublisherPrivate::OnPublishComplete,
```
chapulina: Could use the opportunity to change to `std::bind`

gazebo/transport/Publisher.cc
```diff
@@ -268,6 +308,9 @@ void Publisher::Fini()
     TopicManager::Instance()->Unadvertise(this->topic, this->id);
 
   this->node.reset();
+
+  std::lock_guard<std::mutex> lock(pubMapMutex);
+  publisherMap.erase(this->id);
```
chapulina: I remember having a reason to destroy all publishers before the node, but I can't remember exactly why. Looking through the code, this pattern seems common:

```
this->pub.reset();
this->node->Fini();
this->node.reset();
```

If you don't have a strong reason for terminating the publisher after the node, I'd recommend switching order here.

chapulina (approved):
Wow, nice ABI gymnastics!

> The fix is to bind the callback function along with a shared pointer instead of raw pointer this so that the object is kept alive for a little longer

Interesting, in this case I guess the global static map helped, because if you had used a regular `dataPtr` as `unique_ptr`, it would have died together with the `Publisher`.

gazebo/transport/Publisher.cc
iche033: done. c1e40a5d5a7ad2161c4b1aabbbf57d8430cba2e7

gazebo/transport/Publisher.cc
iche033: moved erase before resetting node c1e40a5d5a7ad2161c4b1aabbbf57d8430cba2e7

iche033:
> Interesting, in this case I guess the global static map helped, because if you had used a regular dataPtr as unique_ptr, it would have died together with the Publisher.

yep the publisher object has to be a shared ptr in a static map. I added a comment in ce2b9df

chapulina (approved):
The latest updates look good too :+1:

PR #2783, opened 2020-07-09
[Gazebo 9] Backport Added ignition common profiler

ahcorde:
This PR is a backport of this other PR https://github.com/osrf/gazebo/pull/2776

Gazebo 9 uses `ign-common1`, in this version of ignition common the profiler is not implemented yet. To backport this functionality into `ign-common` requires to also backport `components` into `ign-cmake0`. I have talked with @chapulina and it makes sense to include the profiler code inside the Gazebo code. In particular I have added this code in `gazebo utils` where the other profiler ( Diagnostic) it's also located.

PR #2813, opened 2020-08-05
Gazebo 9 profiler - take 2

chapulina:
This builds on top of #2783 and includes the changes I suggested on that PR:

* Move the profiler to `gazebo::common` (which is analogous to `ignition::common`)
* Rename all API to `GZ_` so there can't be a collision with Ignition
* Also installed `gz_remotery_vis`

I'm targeting `gazebo9` because I think the diff becomes clearer this way.


CMakeLists.txt
```diff
@@ -37,6 +37,14 @@ if(NOT DEFINED BUILD_TESTING)
     set(BUILD_TESTING OFF)
 endif()
 
+option(ENABLE_PROFILER "Enable Ignition Profiler" FALSE)
```
iche033: `Enable Gazebo Profiler`?

gazebo/common/Profiler.hh
```diff
@@ -0,0 +1,173 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
iche033, line 2: `2018` -> `2020`

gazebo/common/ProfilerImpl.hh
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
iche033, line 2: `2018` -> `2020`

gazebo/common/RemoteryProfilerImpl.cc
```diff
@@ -0,0 +1,244 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
iche033, line 2: `2018` -> `2020`

gazebo/common/RemoteryProfilerImpl.hh
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
iche033, line 2: `2018` -> `2020`

iche033 (commented):
works for me, just a few minor comments. One thing I noticed is that if I don't specify `-DENABLE_PROFILER`, I get a couple of linker errors:

```
libgazebo.so.9.13.2: undefined reference to `gazebo::common::Profiler::~Profiler()'
libgazebo.so.9.13.2: undefined reference to `gazebo::common::Profiler::Profiler()'

```

gazebo/common/Profiler.hh
chapulina: I was leaving them with the original date since they were copied from `ign-common`. I'll add a note to each of them about where they came from, what do you think?

iche033: I see, yeah adding a note sounds good

chapulina: 0afd239

CMakeLists.txt
chapulina: 0afd239

chapulina:
> I get a couple of linker errors:

Fixed the linker errors on 0afd239. The profiler should always be compiled, the difference is whether the macros are actual functions or just void.

iche033:
latest changes look good to me. The gpu-nvidia build failed due to internal compiler error so likely a false positive but the abi-check build has an error about compiling header. Maybe worth trying a rebuild

iche033:

chapulina:
Since we're not installing `Profiler.hh`, it can't be used by the plugins. I'll remove it from there for now.

iche033:
The Examples_build test failed. Looks like we also need to remove the header and macros from the examples.

chapulina:
Argh, I had to remove the profiler from `Event.hh` because that's installed. No profiling for templates :confused: 1b298fc9f2ebf6af0330988752e37c53e9991d52

iche033:
windows build passed :)

iche033 (approved):

PR #2809, opened 2020-07-30
[Gazebo9] Fixed fails for OSX: Added using namespace boost::placeholders

ahcorde:
Related with this issue https://github.com/osrf/gazebo/issues/2808#issuecomment-666036207 and this error in the buildfarm https://build.osrfoundation.org/job/gazebo-ci-pr_any-homebrew-amd64/2057/console

Signed-off-by: Alejandro Hernández <ahcorde@gmail.com>

PR #2816, opened 2020-08-06
gazebo10 -> gazebo11 forward port

j-rivero:


plugins/CMakeLists.txt
```diff
@@ -165,6 +166,10 @@ set (DARTplugins
   GravityCompensationPlugin
 )
 
+set(GAZEBO_PLUGIN_INSTALL_DIR
```
chapulina: Interesting, I don't see this variable used anywhere, and the commit seems to be from 7 years ago

gazebo/rendering/Visual.cc
```diff
@@ -3635,7 +3648,7 @@ void Visual::AddPendingChild(std::pair<VisualType,
 }
 
 /////////////////////////////////////////////////
-void Visual::ProcessMaterialMsg(const msgs::Material &_msg)
+void Visual::ProcessMaterialMsg(const ignition::msgs::Material &_msg)
 {
   if (_msg.has_lighting())
```
chapulina: This line fails with:

```
/var/lib/jenkins/workspace/gazebo-ci-pr_any-ubuntu_auto-amd64-gpu-nvidia/gazebo/gazebo/rendering/Visual.cc:3653:12: error: ���const class ignition::msgs::Material��� has no member named ���has_lighting���; did you mean ���set_lighting���?
   if (_msg.has_lighting())
            ^~~~~~~~~~~~
            set_lighting
```

This is because proto 3 not having optional fields anymore. 

On `gazebo11`, we should use the `gazebo::msgs` version, which users proto 2. See this commit by @scpeters : https://github.com/osrf/gazebo/commit/7d79b37f790f9b076ff9c73b6561787fd3408b37

plugins/CMakeLists.txt
j-rivero: ouch, thanks removed here 580b504

j-rivero:
The [ABI report message](https://build.osrfoundation.org/job/gazebo-abichecker-any_to_any-ubuntu_auto-amd64/620/API_5fABI_20report/):

```
DepthCamera.hh
namespace gazebo::rendering
[+] DepthCamera::CreateNormalsTexture ( std::__cxx11::string const& _textureName )  

| Change | Effect
| -- | --
1st parameter _textureName has been renamed to _normalsName. | No effect.
```
Does not seem like a problem to me.

j-rivero:
Forward port of #2809 to solve problems on compilation on Brew.

gazebo/rendering/Visual_TEST.cc
```diff
@@ -189,7 +189,7 @@ TEST_F(Visual_TEST, BoundingBox)
   ignition::math::AxisAlignedBox emptyBoundingBox = emptyVis->BoundingBox();
   EXPECT_EQ(ignition::math::Vector3d::Zero, emptyBoundingBox.Min());
   EXPECT_EQ(ignition::math::Vector3d::Zero, emptyBoundingBox.Max());
-}
+ }
```
chapulina: Small detail, but this change may trigger the code checker

test/integration/joint_spawn.cc
```diff
@@ -162,7 +162,34 @@ void JointSpawningTest::SpawnJointTypes(const std::string &_physicsEngine,
     }
   }
 
-  // world as child of a joint is not supported in sdf 1.7
```
chapulina: This had been removed on purpose on https://github.com/osrf/gazebo/commit/e4b30950df5ce8cbc8c5f617660d18065e4939ce

iche033:
looks good to me. Waiting for results from jenkins build

chapulina (commented):
Just noticed a couple of things

scpeters:
> The [ABI report message](https://build.osrfoundation.org/job/gazebo-abichecker-any_to_any-ubuntu_auto-amd64/620/API_5fABI_20report/):
> 
> ```
> DepthCamera.hh
> namespace gazebo::rendering
> [+] DepthCamera::CreateNormalsTexture ( std::__cxx11::string const& _textureName )  
> 
> | Change | Effect
> | -- | --
> 1st parameter _textureName has been renamed to _normalsName. | No effect.
> ```
> 
> Does not seem like a problem to me.

the DepthCameraPlugin v-table has also changed, which is a more significant problem, though for a plugin

it also complained about this in the 9 -> 10 forward port: https://github.com/osrf/gazebo/pull/2796#issuecomment-670948761

chapulina:
> the DepthCameraPlugin v-table has also changed, which is a more significant problem, though for a plugin

I suspect that was considered acceptable when merging [this PR](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/3194/page/1) into Gazebo 9, although I didn't find any explicit discussion about it.

`OnNewReflectanceFrame` is pushing down `OnNewNormalsFrame`, but both functions have been recently added and haven't been released into Gazebo 11.0.0, see:

https://github.com/osrf/gazebo/blob/gazebo11_11.0.0/plugins/DepthCameraPlugin.hh

scpeters:
> > the DepthCameraPlugin v-table has also changed, which is a more significant problem, though for a plugin
> 
> I suspect that was considered acceptable when merging [this PR](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/3194/page/1) into Gazebo 9, although I didn't find any explicit discussion about it.
> 
> `OnNewReflectanceFrame` is pushing down `OnNewNormalsFrame`, but both functions have been recently added and haven't been released into Gazebo 11.0.0, see:
> 
> https://github.com/osrf/gazebo/blob/gazebo11_11.0.0/plugins/DepthCameraPlugin.hh

I think we'll just have to rebuild gazebo_plugins:

* https://github.com/ros-simulation/gazebo_ros_pkgs/blob/melodic-devel/gazebo_plugins/include/gazebo_plugins/gazebo_ros_depth_camera.h#L60

chapulina:
> I think we'll just have to rebuild gazebo_plugins:

The ABI report says "_Call of any virtual method at higher position in this class or its subclasses may result in crash or incorrect behavior of applications._"

But no higher virtual methods have ever been released. So I don't think this should break anyone already compiling against Gazebo 11, right?

iche033:
I don't think we've released a new gazebo 11 minor version with the `OnNewNormalsFrame` function right? So it's position relative to `OnNewReflectanceFrame` should be fine as long as the two are at the bottom of the vtable.

scpeters:
> I don't think we've released a new gazebo 11 minor version with the `OnNewNormalsFrame` function right? So it's position relative to `OnNewReflectanceFrame` should be fine as long as the two are at the bottom of the vtable.

ok, thanks, I think that makes it ok.

j-rivero:
All the CI looks good to me. Ready to go.

PR #2819, opened 2020-08-10
[Gazebo 9] Publish performance metrics

ahcorde:
This PR create a new message called `performanceMetrics.proto`. This message contains
 - simulation real time factor
 - sim update sensor rate

Which other field are relevant to publish in this message?

### DIFF
--- a/gazebo/msgs/CMakeLists.txt
+++ b/gazebo/msgs/CMakeLists.txt
@@ -70,6 +70,7 @@ set (msgs
   packet.proto
   physics.proto
   param.proto
+  performance_metrics.proto
   param_v.proto
   planegeom.proto
   pid.proto
--- a/gazebo/msgs/CMakeLists.txt
+++ b/gazebo/msgs/CMakeLists.txt
@@ -8,9 +8,9 @@ set (msgs
   axis.proto
   battery.proto
   boxgeom.proto
-  camerasensor.proto
   camera_cmd.proto
   camera_lens.proto
+  camerasensor.proto
   cessna.proto
   collision.proto
   color.proto
@@ -29,16 +29,16 @@ set (msgs
   geometry.proto
   gps.proto
   gps_sensor.proto
-  gui_camera.proto
   gui.proto
+  gui_camera.proto
   gz_string.proto
   gz_string_v.proto
   header.proto
   heightmapgeom.proto
   hydra.proto
-  imagegeom.proto
   image.proto
   image_stamped.proto
+  imagegeom.proto
   images_stamped.proto
   imu.proto
   imu_sensor.proto
@@ -68,12 +68,12 @@ set (msgs
   model_configuration.proto
   model_v.proto
   packet.proto
-  physics.proto
   param.proto
-  performance_metrics.proto
   param_v.proto
-  planegeom.proto
+  performance_metrics.proto
+  physics.proto
   pid.proto
+  planegeom.proto
   plugin.proto
   pointcloud.proto
   polylinegeom.proto
@@ -84,20 +84,18 @@ set (msgs
   pose_v.proto
   poses_stamped.proto
   projector.proto
-  propagation_particle.proto
   propagation_grid.proto
-  publishers.proto
+  propagation_particle.proto
   publish.proto
+  publishers.proto
   quaternion.proto
-  sonar.proto
-  sonar_stamped.proto
   raysensor.proto
   request.proto
   response.proto
-  rest_response.proto
   rest_login.proto
   rest_logout.proto
   rest_post.proto
+  rest_response.proto
   road.proto
   scene.proto
   selection.proto
@@ -107,6 +105,8 @@ set (msgs
   shadows.proto
   sim_event.proto
   sky.proto
+  sonar.proto
+  sonar_stamped.proto
   spheregeom.proto
   spherical_coordinates.proto
   subscribe.proto
@@ -127,9 +127,9 @@ set (msgs
   wireless_node.proto
   wireless_nodes.proto
   world_control.proto
+  world_modify.proto
   world_reset.proto
   world_stats.proto
-  world_modify.proto
   wrench.proto
   wrench_stamped.proto
 )
--- a/gazebo/msgs/performance_metrics.proto
+++ b/gazebo/msgs/performance_metrics.proto
@@ -22,7 +22,7 @@ message PerformanceMetrics
     required double sim_update_rate  = 3;
 
     /// \brief If the sensor is a camera then this field should be filled
-    /// with fps in real time.
+    /// with average fps in real time.
     optional double fps                     = 4;
   }
 
--- a/gazebo/msgs/performance_metrics.proto
+++ b/gazebo/msgs/performance_metrics.proto
@@ -1,0 +1,21 @@
+syntax = "proto2";
+package gazebo.msgs;
+
+/// \ingroup gazebo_msgs
+/// \interface PerformanceMetrics
+/// \brief A message for run-time simulation performance metrics
+
+message PerformanceMetrics
+{
+
+  message PerformanceSensorMetrics
+  {
+    required string sensor_name             = 1;
+    required double real_sensor_update_rate = 2;
+    required double sim_sensor_update_rate  = 3;
+    optional double fps                     = 4;
+  }
+
+  required double real_time_factor         = 1[default=0.0];
+  repeated PerformanceSensorMetrics sensor = 2;
+}
--- a/gazebo/msgs/performance_metrics.proto
+++ b/gazebo/msgs/performance_metrics.proto
@@ -13,7 +13,7 @@ message PerformanceMetrics
   message PerformanceSensorMetrics
   {
     /// \brief Sensor name
-    required string sensor_name             = 1;
+    required string name             = 1;
 
     /// \brief The update rate of the sensor in real time.
     required double real_update_rate = 2;
--- a/gazebo/msgs/performance_metrics.proto
+++ b/gazebo/msgs/performance_metrics.proto
@@ -16,10 +16,10 @@ message PerformanceMetrics
     required string sensor_name             = 1;
 
     /// \brief The update rate of the sensor in real time.
-    required double real_sensor_update_rate = 2;
+    required double real_update_rate = 2;
 
     /// \brief The update rate of the sensor in simulation time.
-    required double sim_sensor_update_rate  = 3;
+    required double sim_update_rate  = 3;
 
     /// \brief If the sensor is a camera then this field should be filled
     /// with fps in real time.
--- a/gazebo/msgs/performance_metrics.proto
+++ b/gazebo/msgs/performance_metrics.proto
@@ -7,15 +7,31 @@ package gazebo.msgs;
 
 message PerformanceMetrics
 {
-
+  /// \brief This message contains information about the performance of
+  /// each sensor in the world.
+  /// If the sensor is a camera then it will publish the frame per second (fps).
   message PerformanceSensorMetrics
   {
+    /// \brief Sensor name
     required string sensor_name             = 1;
+
+    /// \brief The update rate of the sensor in real time.
     required double real_sensor_update_rate = 2;
+
+    /// \brief The update rate of the sensor in simulation time.
     required double sim_sensor_update_rate  = 3;
+
+    /// \brief If the sensor is a camera then this field should be filled
+    /// with fps in real time.
     optional double fps                     = 4;
   }
 
+  /// max_step_size x real_time_update_rate sets an upper bound of
+  /// real_time_factor. If real_time_factor < 1 the simulation is
+  /// slower than real time.
   required double real_time_factor         = 1[default=0.0];
+
+  /// Each sensor in the world will create a PerformanceSensorMetrics
+  /// message publishing information about the performance.
   repeated PerformanceSensorMetrics sensor = 2;
 }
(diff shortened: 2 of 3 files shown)


gazebo/Server.cc
```diff
@@ -597,7 +707,12 @@ void Server::Run()
 
   this->dataPtr->initialized = true;
 
+  performanceMetricsPub =
+   this->dataPtr->node->Advertise<msgs::PerformanceMetrics>(
+       "~/performance_metrics", 100, 5);
```
iche033: this creates a topic called `/gazebo/gazebo/performance_metrics` for me. Looking at other publishers in this file, we could probably rename this to just `/gazebo/performance_metrics`

gazebo/Server.cc
```diff
@@ -542,6 +548,110 @@ void Server::Fini()
...
+            std::pair<std::string, gazebo::common::Time>(name, 0));
+        worldLastMeasurementTime.insert(
+                std::pair<std::string, gazebo::common::Time>(name, 0));
+        if (ret.second==false)
+        {
+          double updateSimRate =
+            (sensor->LastMeasurementTime() - sensorsLastMeasurementTime[name]).Double();
+          double updateRealRate =
+            (world->RealTime() - worldLastMeasurementTime[name]).Double();
+          sensorsLastMeasurementTime[name] = sensor->LastMeasurementTime();
```
iche033: move this down below since we only need to set the var if `updateSimRate > 0.0`?

gazebo/Server.cc
```diff
@@ -542,6 +548,110 @@ void Server::Fini()
...
+              sensorPerformanceMetrics[name].sensorSimUpdateRate = 1.0/updateSimRate;
+              sensorPerformanceMetrics[name].sensorRealUpdateRate = 1.0/updateRealRate;
+              worldLastMeasurementTime[name] = world->RealTime();
+
+              // Special case for stereo cameras
+              sensors::CameraSensorPtr cameraSensor =
+                std::dynamic_pointer_cast<sensors::CameraSensor>(sensor);
+              if (nullptr != cameraSensor)
+              {
+                sensorPerformanceMetrics[name].sensorFPS = cameraSensor->Camera()->AvgFPS();
```
iche033: The `AvgFPS` data came from ogre and I think it is the **real** time sensor update rate but averaged over a certain window size.

in this case, do we need to publish both the avg and instantaneous `real time sensor update rate`?

gazebo/Server.cc
```diff
@@ -542,6 +548,110 @@ void Server::Fini()
...
+          sensorsLastMeasurementTime[name] = sensor->LastMeasurementTime();
+          if (updateSimRate > 0.0)
+          {
+            struct sensorPerformanceMetricsType emptySensorPerfomanceMetrics;
+            auto ret2 = sensorPerformanceMetrics.insert(
+                std::pair<std::string, struct sensorPerformanceMetricsType>
+                  (name, emptySensorPerfomanceMetrics));
+            if (ret2.second==false)
+            {
+              sensorPerformanceMetrics[name].sensorSimUpdateRate = 1.0/updateSimRate;
```
iche033: I change the physics update rate to 100 instead of 1000 (from gui client left panel: `Physics` -> `real time update rate`), I expect the sensor sim update rate to be the same as what's specified. But with a camera sensor, I see that the sim update rate  is a small value close to 0 and the real update rate is larger than expected. 

It's working correctly for non-rendering sensors though.

gazebo/Server.cc
```diff
@@ -542,6 +548,110 @@ void Server::Fini()
...
+      sensorPerformanceMetric.second.sensorSimUpdateRate);
+    if (sensorPerformanceMetric.second.sensorFPS >= 0.0)
+    {
+      performanceSensorMetricsMsg->set_fps(
+        sensorPerformanceMetric.second.sensorFPS);
+    }
+  }
+
+  // Publish data
+  if (performanceMetricsPub && performanceMetricsPub->HasConnections())
```
iche033: can we move this check to the beginning of the function and return if there are not connections? The down side is that we get one spike in values the first time we publish a msg.

gazebo/Server.cc
ahcorde: The real time factor that is visualized in the Gazebo interface it's using a window of 20 values. The values that I'm publishing are instantaneous. 

is it fine to publish instantaneous values or should I create a small window?

gazebo/Server.cc
ahcorde: I thing that there are two cases:
 - Without lockstep: I think the issue that you are explaining could happen when you are running Gazebo without the lockstep flag. There is no garanties when the callback is called.
 - lockstep: I change the physics update rate to 100 the values remain stable

@iche033  Does it make sense?

gazebo/Server.cc
iche033: looks like there is no specific requirement to publish avg or instantaneous values. One suggestion is to change the `fps` protobuf msg field to be more descriptive of what the values are, e.g. something like `real_avg_fps`

gazebo/Server.cc
iche033: yeah that makes sense. It could be confusing to users that the metrics are correct for some sensors (physics based sensors) but not others (rendering sensors) since they may not be familiar with the internals of gazebo. We may have to have some documentation to explain that the metrics for rendering senors like cameras are more accurate in lock step mode.

iche033: I also just noticed that when I change the real time update rate from 1000 to 100, the `real_tme_factor` values in the published metrics msg takes a while (minutes) to get to the correct factor of 0.1. I think that's because it's computed based on total sim time / real time - so it's giving the overall RTF for the entire simulation. The instantaneous value maybe more useful here.

ahcorde: Added a tutorial in https://github.com/osrf/gazebo_tutorials/pull/117

ahcorde:
Moved logic to `SensorManager.cc`

gazebo/sensors/SensorManager.cc
```diff
@@ -132,6 +136,116 @@ void SensorManager::WaitForSensors(double _clk, double _dt)
...
+  // Real time factor
+  common::Time realTime = world->RealTime();
+  common::Time simTime = world->SimTime();
+  if (realTime == 0)
+    simTime = 0;
+  else
+    simTime = simTime / realTime;
+
+  if (simTime > 0)
+    performanceMetricsMsg.set_real_time_factor(simTime.Double());
```
iche033: reviving the comment about this being the overall RTF for the entire simulation. It maybe more useful to display the current RTF, i.e. instantaneous or avg value over a window size. 

e.g. if I change run the simulation for 5 min and change the `real_time_update_rate` to 100 from 1000, the value published here takes a while to converge down to an RTF of ~0.1

ahcorde: instantaneous RTF: https://github.com/osrf/gazebo/pull/2819/commits/66426aa94c607e97f8f4567143b709119e57fb84

iche033:
I made some changes in fd995c1:

* move variables to top and add comments
* protect the `PublishPerformanceMetrics` function with mutex
* prevent publishing if sim time diff < 0
* style fixes

other than that, the metrics data published look to me

iche033 (approved):

gazebo/sensors/SensorManager.cc
```diff
@@ -132,6 +172,130 @@ void SensorManager::WaitForSensors(double _clk, double _dt)
...
+
+  if (realTime == 0)
+    realTimeFactor = 0;
+  else
+    realTimeFactor = diffSimTime / diffRealtime;
+
+  if (realTimeFactor > 0)
+    performanceMetricsMsg.set_real_time_factor(realTimeFactor.Double());
+  else
+    performanceMetricsMsg.set_real_time_factor(0.0);
```
samuelgundry: Is this if/else necessary?   `realTimeFactor` is either 0 or equal to `realTimeFactor`. 

So isn't this the same as `performanceMetricsMsg.set_real_time_factor(realTimeFactor.Double())`?

gazebo/sensors/SensorManager.cc
```diff
@@ -132,6 +172,130 @@ void SensorManager::WaitForSensors(double _clk, double _dt)
...
+            struct sensorPerformanceMetricsType emptySensorPerfomanceMetrics;
+            auto ret2 = sensorPerformanceMetrics.insert(
+                std::pair<std::string, struct sensorPerformanceMetricsType>
+                  (name, emptySensorPerfomanceMetrics));
+            if (ret2.second == false)
+            {
+              sensorPerformanceMetrics[name].sensorSimUpdateRate =
+                  1.0/updateSimRate;
+              sensorPerformanceMetrics[name].sensorRealUpdateRate =
+                  1.0/updateRealRate;
```
samuelgundry: You can avoid indexing into the map by reusing iterator returned by `insert()`.   Not sure if it'll matter performance wise but it doesn't hurt readability doing so regardless IMO.  Same below for the avg FPS.

gazebo/sensors/SensorManager.cc
```diff
@@ -46,6 +50,42 @@ bool g_sensorsDirty = true;
...
+struct sensorPerformanceMetricsType
+{
+  /// \brief sensor real time update rate
+  double sensorRealUpdateRate;
+
+  /// \brief sensor sim time update rate
+  double sensorSimUpdateRate;
+
+  /// \brief Rendering sensor average real time update rate
+  double sensorFPS;
```
samuelgundry: Call this sensorAvgFPS?

gazebo/sensors/SensorManager.cc
```diff
@@ -46,6 +50,42 @@ bool g_sensorsDirty = true;
...
+/// \brief Data structure for storing metric data to be published
+struct sensorPerformanceMetricsType
+{
+  /// \brief sensor real time update rate
+  double sensorRealUpdateRate;
+
+  /// \brief sensor sim time update rate
+  double sensorSimUpdateRate;
+
+  /// \brief Rendering sensor average real time update rate
```
samuelgundry: Is it worth adding how this `sensorFPS` differs to `sensorRealUpdateRate`? Naively, the naming and this /brief make them sound the same except for one is an average.

gazebo/msgs/performance_metrics.proto
```diff
@@ -0,0 +1,21 @@
...
+
+message PerformanceMetrics
+{
+
+  message PerformanceSensorMetrics
+  {
+    required string sensor_name             = 1;
+    required double real_sensor_update_rate = 2;
+    required double sim_sensor_update_rate  = 3;
+    optional double fps                     = 4;
```
samuelgundry, line 26: This is FPS against the simulation clock, yes, and not wall-clock?  Is it instantaneous FPS (for each tick) or an average since time=0?

gazebo/msgs/performance_metrics.proto
```diff
@@ -0,0 +1,21 @@
...
+/// \brief A message for run-time simulation performance metrics
+
+message PerformanceMetrics
+{
+
+  message PerformanceSensorMetrics
+  {
+    required string sensor_name             = 1;
+    required double real_sensor_update_rate = 2;
+    required double sim_sensor_update_rate  = 3;
```
samuelgundry: depends on your style, but "sensor" seems redundant given the outer message name. And if you keep "Sensor", should you add it to the "fps" field below?

samuelgundry (commented):
Thanks for posting this. Sam from AWS RoboMaker. A couple of clarifying questions w.r.t FPS otherwise looks good. A few comments w.r.t code styles, but these depends on your existing preferences so feel free to ignore/push back.

gazebo/msgs/performance_metrics.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto2";
+package gazebo.msgs;
+
+/// \ingroup gazebo_msgs
+/// \interface PerformanceMetrics
+/// \brief A message for run-time simulation performance metrics
+
+message PerformanceMetrics
```
chapulina, line 8: We should document this message and all of its fields. [Here](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/msgs/camera_lens.proto)'s an example of a documented proto file.

gazebo/msgs/CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ set (msgs
   packet.proto
   physics.proto
   param.proto
+  performance_metrics.proto
```
chapulina: Alphabetize

iche033:
I made some minor tweaks to address a couple unresolved comments. 13f57fe

iche033 (approved):
latest changes look good to me

ahcorde:
@osrf-jenkins retest this please

gazebo/msgs/performance_metrics.proto
```diff
@@ -0,0 +1,37 @@
...
+{
+  /// \brief This message contains information about the performance of
+  /// each sensor in the world.
+  /// If the sensor is a camera then it will publish the frame per second (fps).
+  message PerformanceSensorMetrics
+  {
+    /// \brief Sensor name
+    required string sensor_name             = 1;
```
chapulina: Let's remove `sensor` from this field, because it's redundant

```suggestion
    required string name             = 1;
```

chapulina (commented):
Just one last comment then I think this is good to go!

chapulina (approved):

Issue #2406, opened 2018-01-24
Improve Ignition Fuel support


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/2406)) by Ian Chen (Bitbucket: [Ian Chen](https://bitbucket.org/%7Beaa6fca5-6deb-43f6-907f-971c144735dd%7D/), GitHub: [iche033](https://github.com/iche033)).**

----------------------------------------
a few issues identified in [pull request #2800](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/2800)

* remove the need to set `GAZEBO_MODEL_PATH`

* make sure both server and client is able to download from fuel (currently only client downloads the model files via `InsertModelWidget.cc`)

* test saving world with models from fuel and opening the saved world

* test logging and verify playback on new machine

* support downloading models that contain references to other models

* support  Fuel URIs specified in SDF: `<uri>http://ignitionfuel.org/{owner}/models/{model}</uri>`

PR #2822, opened 2020-08-12
[Gazebo 11] Allow gazebo to download models from Fuel in the sdf files

ahcorde:
Some of the bullet point from the related issue:  https://github.com/osrf/gazebo/issues/2406

> remove the need to set GAZEBO_MODEL_PATH

Models are recorded in `~/.ignition/fuel` I don't need to set up this environment variable. 

> support Fuel URIs specified in SDF

This PR allows to load/download models from Fuel in the sdf files

I create this [world](https://gist.github.com/ahcorde/7e37925e76017327a36d077a936fb26c) to test these features:
 - Model
 - Files (just the mesh, for example dae files)

 > make sure both server and client is able to download from fuel

I used gzserver and the model are being downloaded

> test saving world with models from fuel and opening the saved world

I saved the world with some fuel models. I reopenned the file again and models are shown properly. Some of the URIs are hardcoded to absolute path in my computer

> test logging and verify playback on new machine

If I record a log and then I remove the fuel models from `~/.ignition/fuel` some of the models are not visualize properly (meshes and textures are not being shown). I think It's because how the world is saved, it's using absolute paths.

> support downloading models that contain references to other models

@chapulina can you give some examples of these worlds?


chapulina:
>  It's because how the world is saved, it's using absolute paths.

That should have been fixed in https://github.com/ignitionrobotics/ign-fuel-tools/pull/85. We just need an `ign-fuel-tools4` release to be able to use that on Gazebo 11. I'll work on that soon.

> support downloading models that contain references to other models

After the discussion on https://github.com/ignitionrobotics/ign-fuel-tools/issues/77#issuecomment-648548920, the recommended pattern is using full URLs. And `ign-fuel-tools` should also be handling the `model://` case referring to other assets as discussed here: https://github.com/ignitionrobotics/ign-fuel-tools/pull/85#issue-451587736

ahcorde:
I compiled from sources `ign-fuel-tools4`, right now these three points are working too.

 * > test saving world with models from fuel and opening the saved world
 * > test logging and verify playback on new machine
 * > support downloading models that contain references to other models

ahcorde:
With the last commit we can load worlds from URL using the command line:

```
gazebo --verbose https://staging-fuel.ignitionrobotics.org/1.0/chapulina/worlds/Shapes
```

are worlds fully compatible between Gazebo and Ignition ? I don't know if there any plan to move worlds to fuel too

chapulina:
> are worlds fully compatible between Gazebo and Ignition ? 

We're working on it.

The main pending item now is that Ignition requires a few world plugins to be defined for simulation to work, while classic doesn't need that. This is being addressed by https://github.com/ignitionrobotics/ign-gazebo/pull/281.

After that, the only difference will be some SDF tags here and there, but the worlds should be largely compatible.

ahcorde:
friendly ping @chapulina

gazebo/Server.cc
```diff
@@ -62,6 +62,10 @@
 #include "gazebo/Master.hh"
 #include "gazebo/Server.hh"
 
+#include "ignition/common/Filesystem.hh"
+#include "ignition/common/URI.hh"
+#include "ignition/fuel_tools/Interface.hh"
```
chapulina: Nit: Include Ignition headers with `<>` and move them up, since they're from external projects.

gazebo/common/CommonIface.cc
```diff
@@ -152,7 +154,21 @@ void common::add_search_path_suffix(const std::string &_suffix)
 /////////////////////////////////////////////////
 std::string common::find_file(const std::string &_file)
 {
-  return common::SystemPaths::Instance()->FindFile(_file, true);
+  std::string path = common::SystemPaths::Instance()->FindFile(_file, true);
+  if (path.empty())
+  {
+    gzwarn << "Trying to find the model in Fuel [" << _file << "]" << std::endl;
```
chapulina: I think we could remove the 3 warnings from this function. The function was quiet so far. I can imagine it being used to check if a file exists, in which case it would be spammy to warn the user each time the call fails.

Also, I see `[Wrn] [SystemPaths.cc:459] File or path does not exist [""] [https://staging-fuel.ignitionrobotics.org/1.0/chapulina/worlds/Shapes]` even though the world is loaded correctly. Maybe we can try `SystemPaths` after Fuel, following the pattern you did on `Server.cc`?

gazebo/common/CommonIface.cc
```diff
@@ -152,7 +154,21 @@ void common::add_search_path_suffix(const std::string &_suffix)
...
+    path = ignition::fuel_tools::fetchResource(_file);
```
chapulina: On Gazebo we've been going through the `common::FuelModelDatabase`. I don't remember exactly why now, but I think at least it sets a Gazebo-specific user-agent:

https://github.com/osrf/gazebo/blob/a9f093f2fc7fa7fc49efa73719f6536ad28f8af7/gazebo/common/FuelModelDatabase.cc#L59-L73

What do you think of using that singleton here?

chapulina (commented):
Great work, all the use cases are working for me! Do you think you could add a few tests to make sure we don't break the functionality?

> I create this world to test these features:

Mind committing that world to this repo? I could be called `fuel.world`.

---

I think that with this PR, we can remove the need for the `USE_IGNITION_FUEL` environment variable:

https://github.com/osrf/gazebo/blob/a9f093f2fc7fa7fc49efa73719f6536ad28f8af7/gazebo/gui/InsertModelWidget.cc#L52-L57

gazebo/common/CommonIface.cc
ahcorde: I was trying to use 

```cpp
      path = common::FuelModelDatabase::Instance()->ModelPath(uri);
```

even force the download

```
      path = common::FuelModelDatabase::Instance()->ModelPath(uri, true);
```

But it's not able to download the model.

ahcorde:
Added two tests 
 - check that the model are downloaded and included in the world
 - check world from a URI

gazebo/common/CommonIface.cc
chapulina: Yeah I think you may need to update something inside `FuelModelDatabase`

gazebo/Server.cc
```diff
@@ -402,9 +405,36 @@ bool Server::LoadFile(const std::string &_filename,
...
   }
 
-  if (!sdf::readFile(common::find_file(_filename), sdf))
+  std::string filename = _filename;
+  auto filenameUri = ignition::common::URI(filename);
+
+  if (filenameUri.Scheme() == "http" || filenameUri.Scheme() == "https")
   {
-    gzerr << "Unable to read sdf file[" << _filename << "]\n";
+    std::string downloadedDir = ignition::fuel_tools::fetchResource(filename);
```
chapulina: Can we use `FuelModelDatabase` here too?

gazebo/common/FuelModelDatabase.cc
```diff
@@ -213,26 +213,54 @@ std::string FuelModelDatabase::ModelPath(const std::string &_uri,
   }
```
chapulina, line 219: Now when I open an empty world, I get several warnings like:

```
[Wrn] [FuelModelDatabase.cc:211] URI not supported by Fuel [worlds/empty.world]
[Wrn] [FuelModelDatabase.cc:211] URI not supported by Fuel [worlds/empty.world]
[Wrn] [FuelModelDatabase.cc:211] URI not supported by Fuel [file://media/materials/scripts/gazebo.material]
[Wrn] [FuelModelDatabase.cc:211] URI not supported by Fuel [file://media/materials/scripts/gazebo.material]
```

Since we're using this function to test if Fuel can download a model, we should remove this warning, because the user isn't doing anything wrong.

test/integration/factory.cc
```diff
@@ -1249,6 +1249,68 @@ TEST_F(FactoryTest, FilenameFuelURL)
   // Check model was spawned
   ASSERT_NE(nullptr, world->ModelByName("test_box"));
 }
+
+TEST_F(FactoryTest, worldWithFuelModels)
```
chapulina:
```suggestion
TEST_F(FactoryTest, WorldWithFuelModels)
```

test/integration/factory.cc
```diff
@@ -1249,6 +1249,68 @@ TEST_F(FactoryTest, FilenameFuelURL)
...
+  }
+
+  // Check model was spawned
+  ASSERT_NE(nullptr, world->ModelByName("Radio"));
+}
+
+TEST_F(FactoryTest, FuelURIAsWorldArgument)
+{
+  this->Load(
+    "https://staging-fuel.ignitionrobotics.org/1.0/chapulina/worlds/Shapes",
```
chapulina: I don't think this test is working. For it to run for me, I had to remove the `#ifdef HAVE_IGNITION_FUEL_TOOLS` above (I believe it's a left over from [this PR](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/3139/page/1)).


I made a minor change to fix the test above and uploaded a test world to the OpenRobotics account that has models:

<details><summary>Diff</summary>
<p>

```.diff
diff --git a/test/integration/factory.cc b/test/integration/factory.cc
index f3c1afee3e..27e7903060 100644
--- a/test/integration/factory.cc
+++ b/test/integration/factory.cc
@@ -1222,7 +1222,6 @@ TEST_F(FactoryTest, FilenameModelDatabaseRelativePaths)
 }
 
-#ifdef HAVE_IGNITION_FUEL_TOOLS
 TEST_F(FactoryTest, FilenameFuelURL)
 {
   this->Load("worlds/empty.world", true);
@@ -1233,7 +1232,7 @@ TEST_F(FactoryTest, FilenameFuelURL)
 
   msgs::Factory msg;
   msg.set_sdf_filename(
-      "https://api.ignitionfuel.org/1.0/chapulina/models/Test box");
+      "https://fuel.ignitionrobotics.org/1.0/chapulina/models/Test box");
 
   auto pub = this->node->Advertise<msgs::Factory>("~/factory");
   pub->Publish(msg);
@@ -1250,7 +1249,8 @@ TEST_F(FactoryTest, FilenameFuelURL)
   ASSERT_NE(nullptr, world->ModelByName("test_box"));
 }
 
-TEST_F(FactoryTest, worldWithFuelModels)
+TEST_F(FactoryTest, WorldWithFuelModels)
 {
   this->Load("worlds/fuel_models.world", true);
 
@@ -1270,10 +1270,13 @@ TEST_F(FactoryTest, worldWithFuelModels)
   ASSERT_NE(nullptr, world->ModelByName("Radio"));
 }
 
 TEST_F(FactoryTest, FuelURIAsWorldArgument)
 {
+  // ServerFixture::LoadArgs will split whitespaces, so we use a world name
+  // without spaces
   this->Load(
-    "https://staging-fuel.ignitionrobotics.org/1.0/chapulina/worlds/Shapes",
+    "https://fuel.ignitionrobotics.org/1.0/OpenRobotics/worlds/Test_shapes",
     true);
 
   // World
@@ -1311,7 +1314,6 @@ TEST_F(FactoryTest, FuelURIAsWorldArgument)
   // Check model was spawned
   ASSERT_NE(nullptr, world->ModelByName("sphere"));
 }
-#endif
 
 TEST_P(FactoryTest, InvalidMeshInsertion)
```

</p>
</details>

 And then stopped at this error:

`[Err] [Server.cc:395] Could not open file[https://fuel.ignitionrobotics.org/1.0/OpenRobotics/worlds/Test_shapes]`

It looks like there are more `find_file` functions in `CommonIface` that need to be supported.

ahcorde:
@osrf-jenkins run tests again

gazebo/Server.cc
```diff
@@ -402,9 +405,36 @@ bool Server::LoadFile(const std::string &_filename,
...
+
+        if (fileExtension == "sdf" || fileExtension == "world")
+        {
+          filename = current.c_str();
+        }
+      }
+    }
+  }
+
+  if (!sdf::readFile(common::find_file(filename), sdf))
```
chapulina: You've modified `common::find_file` to look into Fuel. So it's a bit strange to also look into Fuel before calling `common::find_file`. Do we need the extra logic outside this function? What if someone tries to find a Fuel world file with `find_file`? I think they should be able to get it.

gazebo/common/CommonIface.cc
```diff
@@ -159,7 +162,12 @@ void common::add_search_path_suffix(const std::string &_suffix)
...
-  return common::SystemPaths::Instance()->FindFile(_file, true);
+  std::string path = common::FuelModelDatabase::Instance()->ModelPath(_file);
+  if (path.empty())
+  {
+    path = common::SystemPaths::Instance()->FindFile(_file, true);
+  }
+  return path;
 }
 
 /////////////////////////////////////////////////
```
chapulina, line 175: All overloads of the `find_file` function should behave the same way, so I think the functions on line 174 and 180 should also be modified to support Fuel.

gazebo/common/FuelModelDatabase.cc
```diff
@@ -208,31 +210,77 @@ std::string FuelModelDatabase::ModelPath(const std::string &_uri,
...
+              << "      to: " << std::endl
+              << "          " << path << std::endl;
     }
   }
-
-  if (!this->dataPtr->fuelClient->DownloadModel(fuelUri, path))
-  {
-    gzerr << "Unable to download model[" << _uri << "]" << std::endl;
-    return std::string();
+  if (path.empty()) {
```
chapulina, line 246:
```suggestion
  if (path.empty())
  {
```

gazebo/common/FuelModelDatabase.cc
```diff
@@ -208,31 +210,77 @@ std::string FuelModelDatabase::ModelPath(const std::string &_uri,
...
+    }
   }
-  else
-  {
-      gzmsg << "Downloaded model URL: " << std::endl
-            << "          " << _uri << std::endl
-            << "      to: " << std::endl
-            << "          " << path << std::endl;
+
+  // ig path is still empty then try to download a world
```
chapulina: This function is called `ModelPath`, so it's a bit strange to try to download a world here. I think we could create a new function for this.

As for `ModelFile` and `WorldFile`, I don't have a strong preference. I think they could be together with `Model` and `World` respectively.

chapulina (changes requested):

gazebo/Server.cc
ahcorde: updated logic https://github.com/osrf/gazebo/pull/2822/commits/9bad120e6260130351f7f5f86fe334db84b4dad5

gazebo/common/FuelModelDatabase.cc
ahcorde: Added a new method https://github.com/osrf/gazebo/pull/2822/commits/ff1138368bf53d4fd2484fb569e1f135454831be

I was adding everything to the same method to avoid breaking the ABI

gazebo/common/CommonIface.cc
ahcorde: Added some logic https://github.com/osrf/gazebo/pull/2822/commits/176b2df56aaff948475f5d66091e79ac6030e92d

chapulina (commented):
@osrf-jenkins run tests please?

This looks good to me, now we just need to check CI.

chapulina:
@osrf-jenkins run tests maybe?

ahcorde:

chapulina:
It looks like this breaks the Windows build:

```
C:\Jenkins\workspace\gazebo-ci-pr_any-windows7-amd64\ws\install\include\ignition\fuel_tools4\ignition/fuel_tools/Result.hh(48): error C2143: syntax error: missing '}' before '('
```

ahcorde:

chapulina:
@ahcorde , the latest Windows build had the same issue. I vaguely remember seeing issues with fuel-tools enums on Windows. I see that the offending line on CI is [this one](https://github.com/ignitionrobotics/ign-fuel-tools/blob/8bad7f291dc0deb23ef603887df98266c0c541f8/include/ignition/fuel_tools/Result.hh#L48), which has the `DELETE` enum.

ahcorde:

ahcorde:
@chapulina, Fixed windows build, there are some tests failing. The tests that I have edited INTEGRATION_factory it's working.

chapulina (approved):
Functionality works, the new tests pass and it looks like there are no regressions :+1:

PR #2833, opened 2020-08-25
Fix reflectance values for objects without a reflectance map

iche033:
When generating reflectance values for depth cameras, visuals in the scene without a reflectance map is assigned a black material so they show up as 0s in the image output. However, it was found that the material used [Gazebo/Black](https://github.com/osrf/gazebo/blob/e4f20675ef2fbdc8f102adab48365532ff82934f/media/materials/scripts/gazebo.material#L230) is not a pure black color as it has a small amount specular component and the final color is also affected by lighting. So the resulting reflectance output is non-zero. This PR creates a new `Gazebo/BlackUnlit` material that is true black and updates the reflectance generation code to use this material instead.  

### DIFF
--- a/media/materials/scripts/gazebo.material
+++ b/media/materials/scripts/gazebo.material
@@ -240,6 +240,19 @@ material Gazebo/Black
   }
 }
 
+material Gazebo/BlackUnlit
+{
+  technique
+  {
+    lighting off
+    pass
+    {
+      ambient 0 0 0 1
+      diffuse 0 0 0 1
+      specular 0 0 0 1
+    }
+  }
+}
 
 material Gazebo/Red
 {
--- a/gazebo/rendering/DepthCamera.cc
+++ b/gazebo/rendering/DepthCamera.cc
@@ -801,11 +801,11 @@ ReflectanceMaterialListener::ReflectanceMaterialListener(ScenePtr _scene)
   this->reflectanceMaterial =
       Ogre::MaterialManager::getSingleton().getByName(material);
 
-  material = "Gazebo/Black";
+  material = "Gazebo/BlackUnlit";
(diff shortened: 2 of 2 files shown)


ahcorde (approved):

scpeters:
I think this caused a regression in the [OgreLog/LogError](https://github.com/osrf/gazebo/blob/gazebo11_11.1.0/test/integration/ogre_log.cc#L69) test

* https://build.osrfoundation.org/job/gazebo-ci-gazebo9-xenial-amd64-gpu-nvidia/132/testReport/junit/(root)/OgreLog/LogError/

scpeters:
> I think this caused a regression in the [OgreLog/LogError](https://github.com/osrf/gazebo/blob/gazebo11_11.1.0/test/integration/ogre_log.cc#L69) test
> 
> * https://build.osrfoundation.org/job/gazebo-ci-gazebo9-xenial-amd64-gpu-nvidia/132/testReport/junit/(root)/OgreLog/LogError/

~~~
17:22:34: Compiler error: unknown error in gazebo.material(247): token "lighting" is not recognized
~~~

media/materials/scripts/gazebo.material
```diff
@@ -240,6 +240,19 @@ material Gazebo/Black
   }
 }
 
+material Gazebo/BlackUnlit
+{
+  technique
+  {
+    lighting off
```
scpeters, line 247: this line is causing the `INTEGRATION_ogre_log` test to fail:

~~~
17:22:34: Compiler error: unknown error in gazebo.material(247): token "lighting" is not recognized
~~~

iche033: thanks for catching this. Addressed in https://github.com/osrf/gazebo/pull/2840

PR #2840, opened 2020-09-04
Fix ogre log test

iche033:
`lighting` is specified in the wrong place but turns out the reflectance data generated are correct without this flag so it's removed.

### DIFF
--- a/media/materials/scripts/gazebo.material
+++ b/media/materials/scripts/gazebo.material
@@ -244,7 +244,6 @@ material Gazebo/BlackUnlit
 {
   technique
   {
-    lighting off
     pass
     {
       ambient 0 0 0 1


iche033:

iche033:
ogre log test no longer fails in latest [windows build](https://build.osrfoundation.org/job/gazebo-ci-pr_any-windows7-amd64/3004)

j-rivero:
Not sure about the impact of the change in current Gazebo installations. If it is not only used in tests, we might want to add a Changelog entry. Otherwise I don't see a negative effect in the tests.

j-rivero (approved):

iche033:
> Not sure about the impact of the change in current Gazebo installations. If it is not only used in tests, we might want to add a Changelog entry. Otherwise I don't see a negative effect in the tests.

The material was recently added in pull request #2833  and that not been released yet so I think there won't be a negative impact with this change.

PR #2837, opened 2020-08-31
[Gazebo 11] Added profiler to gazebo::rendering and gzclient

ahcorde:
Added profiler to gazebo::rendering

Selección_054

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #2849, opened 2020-09-30
Profile gzclient

chapulina:
Add profiling for `gzclient`, builds on top of #2837 

Set the `RMT_PORT` variable so the `gzserver` and `gzclient` processes go to different ports. I'm using the [same numbers](https://github.com/ignitionrobotics/ign-gazebo/blob/fac36a4b06c3ba45b14c72163f9b390721edbda6/src/cmd/cmdgazebo.rb.in#L400-L418) as `ign-gazebo`, which are 1500 for the server and 1501 for the client. The tutorial needs to be updated if this goes in.

I also added some more profiling points. Here's a run with a camera, a lidar and contact visualization on.

gzclient_profiling

### DIFF
--- a/gazebo/gui/GLWidget.cc
+++ b/gazebo/gui/GLWidget.cc
@@ -19,6 +19,7 @@
 #include <boost/lexical_cast.hpp>
 #include <math.h>
 
+#include <ignition/common/Profiler.hh>
 #include <ignition/math/Matrix4.hh>
 #include <ignition/math/Pose3.hh>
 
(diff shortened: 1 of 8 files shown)


iche033:
nice, I can see the profiling points in remotery now. [Windows build](https://build.osrfoundation.org/job/gazebo-ci-pr_any-windows7-amd64/3025/) caught a compile error and I pushed a fix in 0e5e052. Changes look good to me. I'll wait for a new round of jenkins builds.

ahcorde (approved):

chapulina:
I'll merge this into #2837 so we can get a last round of CI

PR #2853, opened 2020-10-01
Bugfix on the color channel of point clouds from DepthCamera

### LINKED ISSUE #1865: rendering::DepthCamera does not output colors properly for an RGB PointCloud.
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/1865)) by Paul D'Angio (Bitbucket: [pcdangio](https://bitbucket.org/%7B7e4eaee6-2565-46c4-b220-e8c5d4fd600b%7D/)).**

The DepthCamera::PostRender() method of the rendering::DepthCamera class in gazebo/rendering/DepthCamera.cc does not output proper RGB colors for each point.  In the PostRender method, if the <depth_camera>/<output> SDF element is set to "points", it creates a float array containing an RGB point cloud and outputs it through the newRGBPointCloud event.  The format of the float array is X, Y, Z, RGB repeated for all points. (pixels really).  The X, Y, and Z floats for all pixels are fine, but the RGB float for all pixels is incorrect.  The RGB float value always comes out to either:

0x3F800000 if no object is seen by the point (aka X = Y = Z = 0)

0x4AA00000 otherwise, no matter what color the object is that is being seen by that point

I took a look at the gazebo_ros_depth_camera.cc plugin source code written by John Hsu for gazebo_ros_pkgs.  I can tell he had the exact same problem here in trying to use the newRGBPointCloud event, and was forced to instead make a computationally expensive workaround by calculating the RGB point cloud from a depth frame after a newDepthFrame event.  You can even see his attempts at debugging the RGB float problem in his "OnNewRGBPointCloud" method, which was abandoned but never removed from the source code.

The same behavior is made evident [in this post](http://answers.gazebosim.org/question/5469/does-gazebo_ros_depth_camera-with-points-output-produce-a-color-point-cloud/).

GoncaloLeao:
Fixes #1865.

This PR fixes a bug on the fragment shader associated with the depth camera so that the color of each point in the point cloud is correct. 

In order to test this in ROS, the depth camera plugin (which extends Gazebo's `DepthCameraPlugin` class) must correctly extract the colors from the `const unsigned char *_image` on the callback function `virtual void OnNewImageFrame(const unsigned char *_image, unsigned int _width, unsigned int _height, unsigned int _depth, const std::string &_format)`. I created a fork of the gazebo_ros_pkgs repository where I corrected the `GazeboRosDepthCamera` class (https://github.com/GoncaloLeao/gazebo_ros_pkgs/blob/bugfix-rgb-on-depth-camera-cloud/gazebo_plugins/src/gazebo_ros_depth_camera.cpp). When this pull request is accepted, I will create a PR to correct the gazebo_ros_depth_camera plugin of the gazebo_ros_pkgs repository.

In order to test this bugfix, you can use the files I provide in attachment.
1. Download the source code for the gazebo_ros_depth_camera plugin from my fork of gazebo_ros_pkgs, from a branch named 'bugfix-rgb-on-depth-camera-cloud' (https://github.com/GoncaloLeao/gazebo_ros_pkgs/tree/bugfix-rgb-on-depth-camera-cloud).
2. Compile the depth camera plugin.
3. Add the Gazebo Model with a Depth Camera Plugin to your Gazebo models directory. The plugin to be used is the modified version of gazebo_ros_depth_camera. 

[test_sensor.zip](https://github.com/osrf/gazebo/files/5314486/test_sensor.zip) contains all the files needed for the model. It is a modified version of the model of this tutorial: http://gazebosim.org/tutorials/?tut=ros_depth_camera
4. Run Gazebo with any world of your choice and add the depth camera model to the world.
5. On RViz, pay close attention to the colors of the point cloud being published on the topic /camera/depth/points. The colors should closely match the ones for the image topic /camera/color/image_raw.

Here I provide two screenshots of what I can see in Gazebo ...
gazebo_screenshot

... and RViz:

rviz_screenshot


iche033:
I wrote a test for this change in https://github.com/osrf/gazebo/commit/7bca2ed204ac2f3c0d398476c9ebf9b425010335. It places a green box in front of the camera and expects that for every point cloud that has a z value equal to the distance of the box, the color value should be some shade of green. It's currently passing. One weird issue is that when no objects are detected, i.e. z > box distance, the blue pixel value after unpacking the color component always returns 1 (small as it's out of 255). I think that's minor since  the color component should not be used in that case. Can you take a look and apply the test to your branch if it looks good to you?

We should also add some documentation here on how to unpack the floating point color value into r, g, b. One place to add the documentation is [here](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/rendering/DepthCamera.hh#L106)

GoncaloLeao:
Hello @iche033, I ran the test in my local branch and it passed. I applied the change to the `DepthCameraSensorTest` class to my pull request. I am not sure why the blue pixel returns 1 and not 0 when no object are detected.

Regarding the documentation, I agree that it should be made clear how the RGB values can be extracted from the 4th channel. In my last commit, I added some [comments](https://github.com/osrf/gazebo/pull/2853/commits/a117f33193caad6a7bc621d1330e0e3438ab8b88#diff-0c63a2c5c8685f4e06ed35b22cd0376dR107) above the function declaration you suggested. At first, I pondered explaining how to unpack with text only, but I ended up adding a small fragment of code to the comment. This will make it much clearer and remove ambiguity for anyone using the API. What do you think? Is it too much verbose? If so, feel free to make a suggestion for improvement.

gazebo/rendering/DepthCamera.hh
```diff
@@ -104,6 +104,14 @@ namespace gazebo
           unsigned int, const std::string &)>  _subscriber);
 
       /// \brief Connect a to the new rgb point cloud signal
+      /// Point coordinates and color are stored in a vector4f.
+      /// The first three channels are for the XYZ coordinates.
+      /// RGB is packed into the 4th channel as a float.
+      /// Unpacking can be performed as follows:
+      ///   uint8_t red = floor(_pcd[4 * index + 3] / 256.0f / 256.0f);
+      ///   uint8_t green = floor((_pcd[4 * index + 3] - red * 256.0f * 256.0f) / 256.0f);
+      ///   uint8_t blue = floor(_pcd[4 * index + 3] - red * 256.0f * 256.0f - green * 256.0f);
```
iche033: minor coding style issue: can you wrap these lines to 80 characters? e.g.

```suggestion
      ///   uint8_t green = floor((_pcd[4 * index + 3] - red * 256.0f * 256.0f)
      ///       / 256.0f);
      ///   uint8_t blue = floor(_pcd[4 * index + 3] - red * 256.0f * 256.0f -
      ///       green * 256.0f);
```

iche033 (commented):
Just a minor coding style suggestion otherwise the changes and the documentation look good.

gazebo/rendering/DepthCamera.hh
GoncaloLeao: Ok, I updated those lines of the documentation.

iche033 (approved):

j-rivero:
@GoncaloLeao  could you please pull changes in gazebo11 branch into your branch to get latest CI before merging?

GoncaloLeao:
I applied the changes from gazebo11.
This branch should now be ready to merge.

chapulina:
@GoncaloLeao , it looks like there are still conflicts preventing this from being merged.

scpeters:
I just resolved conflicts and pushed to restart CI

scpeters:
Thanks for submitting the fix and sorry for the delay in merging! If you are able to make a matching gazebo_ros_pkgs pull request I will be happy to review it.

PR #2856, opened 2020-10-09
Fix gzclient crash when deleting a model that's being manipulated

iche033:
The crash happens because the visual pointer in `ModelManipulator` is not reset when the model is being deleted.

To reproduce:

1. launch gazebo, `gazebo --verbose`
1. insert box model using toolbar
1. On separate terminal, publish a delay msg to delete the box model
    ```
    sleep 5 && gz model -d -m unit_box
    ```
1. Go back to the gazebo window, press 'T' to go into translation mode, and continuously drag and move the box around until it's deleted
1. Release mouse button to see the crash

The fix is to reset the visual pointer when detaching the model. I also had to fix a material transparency issue that happens when I tried to insert the same model back into the world (after being deleted while dragging) and found that It always reverted to use old semi-transparent material. The workaround is to also reset the transparency before resetting the visual pointer. 

### DIFF
--- a/gazebo/gui/ModelManipulator.cc
+++ b/gazebo/gui/ModelManipulator.cc
@@ -106,6 +106,17 @@ void ModelManipulator::Detach()
   this->dataPtr->selectionObj->SetMode(
       rendering::SelectionObj::SELECTION_NONE);
   this->dataPtr->selectionObj->Detach();
+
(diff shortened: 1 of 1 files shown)


j-rivero:

ahcorde (approved):
LGTM, I reproduced the crash and the fix

PR #2863, opened 2020-10-24
Fix physics based sensor update rate in lockstep mode

iche033:
closes issue #2861 - see issue on how to reproduce the problem

Problem was that the physics based sensors were all throttled by the max update rate of all sensors in lockstep mode. This PR applies proper throttling to the sensors.

I was also able to reproduce the problem by adding an additional ray sensor with different update rate to the laser strict rate integration test world file and that caused `INTEGRATION_laser` test to fail. Test should be fixed with these changes.

### DIFF
--- a/gazebo/sensors/Sensor.cc
+++ b/gazebo/sensors/Sensor.cc
@@ -174,8 +174,18 @@ void Sensor::Update(const bool _force)
   {
     if (this->useStrictRate)
     {
+      // rendering sensors (IMAGE category) has it's own mechanism
+      // for throttling and lockstepping with physics. So throttle just
+      // physics sensors
+      if (this->dataPtr->category != IMAGE && !this->NeedsUpdate() && !_force)
+        return;
+
       if (this->UpdateImpl(_force))
+      {
+        common::Time simTime = this->world->SimTime();
+        this->lastUpdateTime = simTime;
         this->dataPtr->updated();
+      }
     }
     else
     {
--- a/gazebo/sensors/Sensor.cc
+++ b/gazebo/sensors/Sensor.cc
@@ -174,7 +174,7 @@ void Sensor::Update(const bool _force)
   {
     if (this->useStrictRate)
     {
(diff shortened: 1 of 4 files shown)


ahcorde (approved):

gazebo/sensors/Sensor.cc
```diff
@@ -174,8 +174,18 @@ void Sensor::Update(const bool _force)
...
+      // rendering sensors (IMAGE category) has it's own mechanism
+      // for throttling and lockstepping with physics. So throttle just
+      // physics sensors
+      if (this->dataPtr->category != IMAGE && !this->NeedsUpdate() && !_force)
+        return;
+
       if (this->UpdateImpl(_force))
+      {
+        common::Time simTime = this->world->SimTime();
+        this->lastUpdateTime = simTime;
```
chapulina: Nit: could avoid the temporary object here:

```suggestion
        this->lastUpdateTime = this->world->SimTime();
```

chapulina (approved):
I haven't tried it manually, but the logic looks reasonable.

gazebo/sensors/Sensor.cc
iche033: done. 4ab78c5

peci1:
This fix is very much needed in Gazebo 11, where it hasn't been ported!!! Without it, gazebo_ros_imu_sensor is broken and never publishes any ROS messages in lockstep mode (and gazebo_ros_block_laser and gazbeo_ros_range as well, as given by the search https://github.com/search?q=repo%3Aros-simulation%2Fgazebo_ros_pkgs%20lastupdatetime&type=code ).

When porting the fix, it should also be taken into account that `lastUpdateTime` is protected by mutex `this->dataPtr->mutexLastUpdateTime`, which wasn't locked in the code added by this PR.

@iche033 Do you want to prepare the forward port or should I?

iche033:
> @iche033 Do you want to prepare the forward port or should I?

yep, created #3353

PR #2871, opened 2020-11-06
Add mutex to make Sensor::SetActive threadsafe

iche033:
The `IsActive` function in the Sensor class is being called from various threads, e.g. [rendering](https://github.com/osrf/gazebo/blob/gazebo9/gazebo/sensors/CameraSensor.cc#L254), [physics](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/sensors/SensorManager.cc#L402) (event callback from physics), and [transport](https://github.com/osrf/gazebo/blob/gazebo9/gazebo/sensors/SonarSensor.cc#L344) threads. So a mutex is added to protects calls for setting the `active` variable in `Sensor` class so that it is safe to call `SetActive` from gazebo plugins, e.g. various plugins in gazebo_ros_pkgs calls [SetActive](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_block_laser.cpp#L195) to enable / disable the sensors when subscriber count changes. 

### DIFF
--- a/gazebo/sensors/ContactSensor.cc
+++ b/gazebo/sensors/ContactSensor.cc
@@ -323,7 +323,7 @@ void ContactSensor::OnContacts(ConstContactsPtr &_msg)
 //////////////////////////////////////////////////
 bool ContactSensor::IsActive() const
 {
-  return this->active ||
+  return Sensor::IsActive() ||
     (this->dataPtr->contactsPub &&
      this->dataPtr->contactsPub->HasConnections());
 }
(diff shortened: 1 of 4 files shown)


ahcorde (approved):
LGTM

chapulina (approved):
LGTM. I'm just wondering if the access to `active` inside `Sensor::Fini` also needs to be protected, since that's a public function.

iche033:
yep added mutex in `Fini` just in case. 59a5f75

Issue #2881, opened 2020-11-18
Dem size is not computed correctly for extraterrestrial celestial bodies


scpeters:
While debugging #2828 (a hard failure of gazebo when loading certain Dem files), I noticed the following error message from the reported `UNIT_Dem_TEST` failure:

~~~
[ RUN      ] DemTest.NegDem
ERROR 1: PROJ: proj_create_operations: Source and target ellipsoid do not belong to the same celestial body
ERROR 6: Cannot find coordinate operations from
 `PROJCRS["Moon2000_npole",BASEGEOGCRS["GCS_Moon",DATUM["Moon_2000",ELLIPSOID["Moon_2000_IAU_IAG",1737400,0,LENGTHUNIT["metre",1,ID["EPSG",9001]]]],PRIMEM["Reference_Meridian",0,ANGLEUNIT["degree",0.0174532925199433,ID["EPSG",9122]]]],CONVERSION["unnamed",METHOD["Polar Stereographic (variant B)",ID["EPSG",9829]],PARAMETER["Latitude of standard parallel",90,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8832]],PARAMETER["Longitude of origin",0,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8833]],PARAMETER["False easting",0,LENGTHUNIT["metre",1],ID["EPSG",8806]],PARAMETER["False northing",0,LENGTHUNIT["metre",1],ID["EPSG",8807]]],CS[Cartesian,2],AXIS["easting",south,ORDER[1],LENGTHUNIT["metre",1]],AXIS["northing",south,ORDER[2],LENGTHUNIT["metre",1]]]'
 to `EPSG:4326'
~~~

Based on the first error message "Source and target ellipsoid do not belong to the same celestial body" and noting that [EPSG:4326](https://epsg.io/4326) refers to the WGS84 model of earth and that the `PROJCRS` string for [dem_neg.tif](https://github.com/osrf/gazebo/blob/gazebo11/test/media/materials/textures/dem_neg.tif) used in this test appears to refer to the moon, there are clearly issues with the coordinate transformations used in our `Dem` class.

The following is the output of `gdalinfo dem_neg.tif -mm`:

~~~
Driver: GTiff/GeoTIFF
Files: ../test/media/materials/textures/dem_neg.tif
Size is 33, 33
Coordinate System is:
PROJCS["Moon2000_npole",
    GEOGCS["GCS_Moon",
[...]

Issue #2828, opened 2020-08-22
UNIT_Dem_TEST failure on 20.04


richmattes:
# Description
The UNIT_Dem_TEST fails for me with a segfault, with the following output:
```
        Start  19: UNIT_Dem_TEST
 19/424 Test  #19: UNIT_Dem_TEST .........................................***Exception: SegFault  0.46 sec
[==========] Running 10 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 10 tests from DemTest
[ RUN      ] DemTest.MisingFile
[Wrn] [SystemPaths.cc:459] File or path does not exist ["/file/shouldn/never/exist.png"] [/file/shouldn/never/exist.png]
[Err] [Dem.cc:78] Unable to open DEM file[/file/shouldn/never/exist.png], check your GAZEBO_RESOURCE_PATH settings.
[       OK ] DemTest.MisingFile (2 ms)
[ RUN      ] DemTest.NotDem
ERROR 4: `/home/rich/fedora/gazebo/gazebo-11.1.0/media/materials/scripts/CMakeLists.txt' not recognized as a supported file format.
[Err] [Dem.cc:88] Unable to open DEM file[/home/rich/fedora/gazebo/gazebo-11.1.0/media/materials/scripts/CMakeLists.txt]. Format not recognised as a supported dataset.
[       OK ] DemTest.NotDem (0 ms)
[ RUN      ] DemTest.UnsupportedDem
[Err] [Dem.cc:96] Unsupported number of bands in file [/home/rich/fedora/gazebo/gazebo-11.1.0/media/materials/textures/wood.jpg]. Found 3 but only 1 is a valid value.
[       OK ] DemTest.UnsupportedDem (1 ms)
[ RUN      ] DemTest.NonSquaredDemPortrait
[       OK ] DemTest.NonSquaredDemPortrait (24 ms)
[ RUN      ] DemTest.NonSquaredDemLandscape
[       OK ] DemTest.NonSquaredDemLandscape (12 ms)
[ RUN      ] DemTest.SquaredDem
[       OK ] DemTest.SquaredDem (11 ms)
[ RUN      ] DemTest.BasicAPI
[Err] [Dem.cc:181] EXCEPTION: Illegal coordinates. You are asking for the elevation in (0,129) but the terrain is [129 x 129]


[Err] [Dem.cc:181] EXCEPTION: Illegal coordinates. You are asking for the elevation in (129,0) but the terrain is [129 x 129]


[...]

PR #2882, opened 2020-11-19
Warn instead of fail for non-Earth Dem's on 20.04 (#2828)

scpeters:
A test failure related to the `gazebo::common::Dem` class on Ubuntu 20.04 was reported in #2828 and seems to be caused by loading Dem's that are based on a celestial body other than earth. The code doesn't fail like this in 18.04 (though I believe it computes Dem sizes incorrectly as reported in #2881) but it is now failing in 20.04 by returning a `nullptr` from `OGRCreateCoordinateTransformation` if the source and target coordinate reference systems refer to different celestial bodies. Since `Dem::GetGeoReference` returns `void` instead of `bool`, I've added a `gzthrow` to that function if the coordinate transformation is unsuccessful along with try / catch blocks that print `gzwarn` console messages in the two places that call this API:

1. `Dem::Load` calls `Dem::GetGeoReference` for the corners of the `Dem` to compute the size
2. `HeightmapShape::LoadTerrainFile` calls `Dem::GetGeoReferenceOrigin` in order to configure the `SphericalCoordinates` class used by the gazebo gui and the `GpsSensor`

The `UNIT_Dem_TEST` will still fail on 20.04 with these changes, but at least it won't crash.


jacobperron (approved):
LGTM

gazebo/physics/HeightmapShape.cc
```diff
@@ -116,7 +116,16 @@ int HeightmapShape::LoadTerrainFile(const std::string &_filename)
       ignition::math::Angle latitude, longitude;
       double elevation;
 
-      this->dem.GetGeoReferenceOrigin(latitude, longitude);
+      try
+      {
+        this->dem.GetGeoReferenceOrigin(latitude, longitude);
+      }
+      catch(const common::Exception &_e)
```
chapulina: `_e` isn't used, I think it could be omitted, or you could print its message.

gazebo/common/Dem.cc
```diff
@@ -106,25 +106,34 @@ int Dem::Load(const std::string &_filename)
...
+
+    // Set the world width and height
+    this->dataPtr->worldWidth =
+       common::SphericalCoordinates::Distance(upLeftLat, upLeftLong,
+                                              upRightLat, upRightLong);
+    this->dataPtr->worldHeight =
+       common::SphericalCoordinates::Distance(upLeftLat, upLeftLong,
+                                              lowLeftLat, lowLeftLong);
+  }
+  catch(const common::Exception &)
```
chapulina, line 131: Consider printing the exception's message.

chapulina (approved):
Looks good to me, I just have a couple of suggestions about the exception messages. Also, we should consider documenting that some functions may throw.

gazebo/physics/HeightmapShape.cc
scpeters: I'll remove `_e`; I had been intending to print the exception message, but then I realized that `gzthrow` already prints the message:

* [gzthrow creates an Exception with the multi-argument constructor](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/common/Exception.hh#L39-L41)
* [that constructor calls `Exception::Print`](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/common/Exception.cc#L36-L41)

chapulina: Oh nice :+1:

scpeters: removed `_e` in d217b05

PR #2885, opened 2020-11-20
Update TinyOBJLoader to v2.0.0rc8

### LINKED ISSUE #2870: Support filenames containing spaces
File names and paths containing spaces have never been officially supported by Gazebo classic. There are use cases which are known to work, but because this has never been a well supported use case, there are corner cases that won't work, and there's also the possibility that new functionality breaks existing cases that happen to work.

This issue is about making the support for spaces in file names and paths official. Test cases will be added for all supported file types and ways of inserting them into simulation, to make sure they continue to be supported. In the process, the use cases which don't currently work will be fixed.

The following file types will support spaces:

* World files
    - [x] From the command line: `gazebo <world file>`
* Model files
    - [x] Through the `~/factory` topic
    - [x] Through the `<include>` SDF tag
    - [x] From the GUI
* Mesh files (for visuals and collisions, DAE, OBJ, STL)
    - [x] Referenced in SDF files
* Material files
    - [x] Ogre material scrips referenced in SDF files
    - [x] MTL files referenced in OBJ files (not supported, see https://github.com/tinyobjloader/tinyobjloader/issues/46)
* Normal map file
    - [x] Referenced in SDF files
* Actor skins and animations
    - [x] Referenced in SDF files
* Textures
    - [x] Referenced in COLLADA files (#2369)
    - [x] Referenced in MTL files
    - [x] Referenced in material scripts

Notably, plugin files (shared libraries) are not included in this list.

chapulina:
Builds on top of #2877, resolves #2870. 

This completes the support of files containing white spaces. The support for referring to MTL files with spaces from OBJ files has been added to TinyOBJLoader on https://github.com/tinyobjloader/tinyobjloader/pull/285 and is staged to be released into their 2.0.0 version.

I don't think it's a great idea to use a release candidate, but it looks like the `v2.0.0rc8` tag doesn't break any of our existing use cases and brings lots of new features.

The only API change from the revision we were previously using was the addition of warning messages, and that was added back in `v1.3.0`, so it's not an artifact of bumping the major version.

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -5,6 +5,13 @@ Deprecated code produces compile-time warnings. These warning serve as
 notification to users that their code should be upgraded. The next major
 release will remove the deprecated code.
 
+## Gazebo 9.15 to 9.16
+
+### Modifications
+
+Updated the version of TinyOBJLoader from 1.0.0 to 2.0.0rc8.
+See the changelog at https://github.com/osrf/gazebo/blob/gazebo9/deps/tinyobjloader/tiny_obj_loader.h
+
 ## Gazebo 8.4 to 9.x
 
 ### Models with duplicate names will not be inserted
--- a/test/models/testdb/model with spaces/meshes/material with spaces.mtl
+++ b/test/models/testdb/model with spaces/meshes/material with spaces.mtl
@@ -1,8 +1,7 @@
-newmtl everything is ignored after first space
+newmtl MaterialNameWithoutSpaces
   Ns 10.0
   Ni 1.5
   d 1.0
-  Tr 0.0
   Tf 1.0 1.0 1.0
   illum 2
   Ka 1.0 1.0 1.0
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,12 @@
 
 ## Gazebo 9.xx.x (202x-xx-xx)
 
+1. Updated the version of TinyOBJLoader from 1.0.0 to 2.0.0rc8.
+    * [Pull request #2885](https://github.com/osrf/gazebo/pull/2885)
+
+1. Support resource files with spaces
+    * [Pull request #2877](https://github.com/osrf/gazebo/pull/2877)
+
 1. Fix physics based sensor update rate in lockstep mode
     * [Pull request #2863](https://github.com/osrf/gazebo/pull/2863)
 
--- a/deps/tinyobjloader/tiny_obj_loader.h
+++ b/deps/tinyobjloader/tiny_obj_loader.h
@@ -1,7 +1,7 @@
 /*
 The MIT License (MIT)
 
-Copyright (c) 2012-2016 Syoyo Fujita and many contributors.
+Copyright (c) 2012-2018 Syoyo Fujita and many contributors.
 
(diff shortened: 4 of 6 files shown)


test/models/testdb/model with spaces/meshes/material with spaces.mtl
```diff
@@ -1,8 +1,7 @@
-newmtl everything is ignored after first space
+newmtl MaterialNameWithoutSpaces
   Ns 10.0
   Ni 1.5
   d 1.0
-  Tr 0.0
```
j-rivero, line 5: the change was motivated by any particular error/problem? change in the spec? or just because it is the default?

Migration.md
```diff
@@ -5,6 +5,12 @@ Deprecated code produces compile-time warnings. These warning serve as
 notification to users that their code should be upgraded. The next major
 release will remove the deprecated code.
 
+## Gazebo 9.15 to 9.16
+
+### Modifications
+
+Updated the version of TinyOBJLoader from 1.0.0 to 2.0.0rc8.
```
j-rivero, line 12: I see that the API seems to be working after the bump in major version. Not sure if there is any behavior change or anything technical worth to mention here.

j-rivero (commented):
Did not find anything wrong in the PR or in the TinyOBJLoader bump. We can update changelog in this PR (unless you plan to update it with the changes for making the files with spaces to work):

```diff
diff --git a/Changelog.md b/Changelog.md
index 37a5d4c38a..cb80e23f18 100644
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,9 @@
 1. Fix physics based sensor update rate in lockstep mode
     * [Pull request #2863](https://github.com/osrf/gazebo/pull/2863)
 
+1. Updated the version of TinyOBJLoader from 1.0.0 to 2.0.0rc8. 
+    * [Pull request #2885](https://github.com/osrf/gazebo/pull/2885)
+
 ## Gazebo 9.15.0 (2020-09-30)
 
 1. More enhancement for Windows build
```
I'm compiling the beast to give it a quick try before approve it.

test/models/testdb/model with spaces/meshes/material with spaces.mtl
chapulina: Keeping this here now gives warnings that "Both `d` and `Tr` parameters defined ... Use the value of `d` for dissolve". That was changed since version 1.0.5 and refers to this issue: https://github.com/tinyobjloader/tinyobjloader/issues/43.

Migration.md
chapulina: Added a link to the changelog

chapulina:
Updated the changelog and migration. Also rebased now that #2877 was merged - I hope that doesn't mess up your workspace, @j-rivero .

j-rivero:
> Updated the changelog and migration. Also rebased now that #2877 was merged - I hope that doesn't mess up your workspace, @j-rivero .

No problem. I did not find any problem on the runtime manual test. All the failing tests seems not related to the ones involved in this PR, specially factory and obj_loader. Nice work.

j-rivero (approved):

PR #2888, opened 2020-11-23
[Forward port] Support spaces in resource files, upgrade TinyOBJLoader

chapulina:
Ports #2877 and #2885 from `gazebo9` to `gazebo11`.


j-rivero (commented):
Changelog and migration has been forwarded for gazebo9, we probably want the same for gazebo11. I'm waiting for the CI and local compilation, changes look good to me.

chapulina:
> Changelog and migration has been forwarded for gazebo9, we probably want the same for gazebo11

Good point. How about I add those when I open a PR to bump the version and make the release? To save on CI. I can do that as soon as this is merged.

j-rivero:
> Good point. How about I add those when I open a PR to bump the version and make the release? To save on CI. I can do that as soon as this is merged.

perfect. My compilation won't finish before I leave the work by today. If you need to merge, I think it's fine to do it with a clear CI on affected tests. I can comment afterwards if I find something locally.

j-rivero:
@osrf-jenkins run tests please . We lost them after a bug in the buildfarm.

j-rivero:
Can not find any problem in manual testing. Let's wait a little bit for the Linux/Mac results on CI, but I think this mostly ready to go.

j-rivero (approved):

PR #2892, opened 2020-11-27
Fix namespace usage in console macros

stefanbuettner:
Hey there,

This code did not work
```cpp
namespace foo::gazebo
{

class bar
{
public:
    void doSomething() {
        gzmsg << "doSomething" << std::endl;
    }
};

}  // foo::gazebo
```

because gzmsg resolves to (gazebo::common::Console::msg()) which
resolves in this error:
```
error: ‘foo::gazebo::common’ has not been declared
     #define gzmsg (gazebo::common::Console::msg())
```

Should this also be fixed in older gazebo version?

This solves #2896 

Cheers,
Stefan

cc @Pro

### DIFF
--- a/gazebo/common/Console.hh
+++ b/gazebo/common/Console.hh
@@ -38,19 +38,19 @@ namespace gazebo
     /// \{
 
     /// \brief Output a message
-    #define gzmsg (gazebo::common::Console::msg())
+    #define gzmsg (::gazebo::common::Console::msg())
 
     /// \brief Output a debug message
-    #define gzdbg (gazebo::common::Console::dbg(__FILE__, __LINE__))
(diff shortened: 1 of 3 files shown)


chapulina:
Interesting, thank you for the fix. Do you think you could add a test case that wouldn't compile without the fix?

> Should this also be fixed in older gazebo version?

Yeah that would be great. We can backport this to `gazebo9` after it's merged. I wouldn't worry about Gazebo 7 and 10 since they only have 2 months until EOL.

stefanbuettner:
Yes, I'll add a regression test. I also opened up an issue as a test reference.

scpeters (approved):
thanks for the test and bug fix! you rock!

Issue #800, opened 2013-08-09
Plugin loading on OS X


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/800)) by William Woodall (Bitbucket: [William Woodall](https://bitbucket.org/%7B111b81a3-502e-44d1-9cf6-c11de0e697cb%7D/), GitHub: [wjwwood](https://github.com/wjwwood)).**

----------------------------------------
There are many places where the `.so` extension is assumed through out gazebo and the related packages.

For example, when loading a plugin with the `-s` command line option it is recommended that users give `libfoo.so`. On OS X this will not work because the file name is likely called `libfoo.dylib`. This is a problem in the `gazebo_ros` package for example:

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_ros/scripts/gazebo#L7

Other places include specifying plugins in the urdf description of robots, for example:

https://github.com/davetcoleman/baxter_common/blob/development/baxter_description/urdf/baxter.urdf#L1230

Any time `.so` is hardcoded or assumed there will be a break on OS X.

In ROS we use pluginlib (which uses class_loader):

https://github.com/ros/pluginlib
https://github.com/ros/class_loader

Pluginlib works around this problem by only specifying the first part of the library name, leaving off the suffix. Then pluginlib looks for the correct thing based on the OS:

See the path attribute:
http://ros.org/wiki/pluginlib#pluginlib.2BAC8-pluginlib_groovy.The_Plugin_Description_File

I am creating this issue to help track what needs to be done to become agnostic to the dynamic library extension being used.

PR #2897, opened 2020-12-04
Compile test plugins as MODULE instead of STATIC

### LINKED ISSUE #2883: PluginTest failing since plugins became static libraries
These tests started failing recently:

* `PluginTest.ModelExceptionConstructor`
* `PluginTest.ModelExceptionInit`
* `PluginTest.ModelExceptionLoad`

I can reproduce the failures locally. The problem is that the plugins can't be loaded anymore. Here's the test log:

```
[Err] [Plugin.hh:178] Failed to load plugin libExceptionModelPluginConstructor.so: libExceptionModelPluginConstructor.so: cannot open shared
 object file: No such file or directory                                                                                                     
...
/home/chapulina/gazebo11_ws/src/gazebo/test/integration/plugin.cc:58: Failure                                  
Value of: loggedString.find( "Exception occured in the constructor of plugin with name") != std::string::npos
  Actual: false                                                                         
Expected: true
```

Changing the libraries to `SHARED` here fixes the tests for me:

https://github.com/osrf/gazebo/blob/5ecd18aead9cf70574486979b161ad5c12a5faa2/test/plugins/CMakeLists.txt#L34

This was changed to `STATIC` on #2753 and #2754.

@j-rivero , do you think we can revert that one change? Or maybe fix the test some other way? Right now the plugins have `.a` extensions and the test is expecting `.so`.

j-rivero:
Revert part of #2753 which change the compilation of plugins to STATIC.

Should fix #2883 in the gazebo11 branch. Needs to be tested on ARM.


scpeters:
we still need to test with ARM, but just looking at our other supported platforms, I believe this has fixed the tests on Ubuntu amd64, but the macOS tests are still broken. apparently `MODULE` libraries have a `.so` suffix on macOS (while shared libraries end with `.dylib`, see #800 and this [stackover answer](https://stackoverflow.com/a/2339910)), so I think we'll need to update the logic in [Plugin.hh](https://github.com/osrf/gazebo/blob/6fd426b3949c4ca73fa126cde68f5cc4a59522eb/gazebo/common/Plugin.hh#L126-L134) in order to load `.so` files as plugins on macOS

~~~
[ 53%] Linking CXX shared module libExceptionModelPluginConstructor.so
...
[ 53%] Linking CXX shared module libExceptionModelPluginInit.so
~~~

we don't need to block this PR on macOS; if it fixes Ubuntu amd64 and ARM, we can add some comments to #800 about this and follow up later

j-rivero:
> Needs to be tested on ARM.

Let's try [Build Status](https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/)

> apparently `MODULE` libraries have a `.so` suffix on macOS (while shared libraries end with `.dylib`, see #800 and this [stackover answer](https://stackoverflow.com/a/2339910)), so I think we'll need to update the logic in [Plugin.hh](https://github.com/osrf/gazebo/blob/6fd426b3949c4ca73fa126cde68f5cc4a59522eb/gazebo/common/Plugin.hh#L126-L134) in order to load `.so` files as plugins on macOS

Interesting. I agree on following on #800.

scpeters:
@j-rivero I just merged with `gazebo11`, so this should be up to date now. Do you think we can try another ARM test build of this branch?

scpeters:
arm test build:

* [Build Status](https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/3/) https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/3/

scpeters:
> arm test build:
> 
> * [Build Status](https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/3/) https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/3/

I started another job but it didn't finish

[Build Status](https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/4/) https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/4/

scpeters:
it's looking better; the arm build succeeded, though there are more test failures than normal and the console output suggests a runtime problem

~~~
395: Unsupported setsockopt level=0 optname=32
395: Error setting socket option (IP_MULTICAST_IF).
395: Unsupported setsockopt level=0 optname=32
395: Error setting socket option (IP_MULTICAST_IF).
395: Did you set the environment variable IGN_IP with a correct IP address? 
395:   [172.17.0.2] seems an invalid local IP address.
395:   Using 127.0.0.1 as hostname.
395: terminate called after throwing an instance of 'std::out_of_range'
395:   what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
395: qemu: uncaught target signal 6 (Aborted) - core dumped
~~~

[Build Status](https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/7/) https://build.osrfoundation.org/job/_test_gazebo-ci-pr_any-ubuntu_auto-arm64-gpu-nvidia/7/

the windows build is failing though

~~~
Scanning dependencies of target gtest_main
Error: dependent 'test\gtest.lib' does not exist.
~~~

[Build Status](https://build.osrfoundation.org/job/gazebo-ci-pr_any-windows7-amd64/3203/) https://build.osrfoundation.org/job/gazebo-ci-pr_any-windows7-amd64/3203/

scpeters (approved):
thanks for all the iteration and fixing the changelog!

Issue #2584, opened 2019-01-15
Incorrect illumination in camera sensor images


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/2584)) by Christian Rauch (Bitbucket: [christian-rauch](https://bitbucket.org/%7B408871b1-9dca-4a44-88df-e8f0dbdba533%7D/)).**

The original report had attachments: [camera_lighting_problem.world](https://osrf-migration.github.io/gazebo-gh-pages/data/repositories/osrf/gazebo/issues/2584/attachments/camera_lighting_problem.world)

----------------------------------------
In a minimal example world with a **camera** and a **depth** sensor (see attached world file), the colour images have wrong illumination:
![gazebo_camera_illumination_problem.png](https://osrf-migration.github.io/gazebo-gh-pages/data/bitbucket.org/repo/jgXqbo/images/3851408910-gazebo_camera_illumination_problem.png)

While the visualised image (the one in the camera frustum) has the correct colours, the published images are missing the correct illumination. E.g. they are missing shadows and the colours are less vivid.

If *always_on* is disabled for the depth sensor, it sometimes seem to work.

PR #2907, opened 2020-12-15
Fix depth camera breaking shadows

Benlitz:
In the code of the depth camera, `setShadowTechnique(Ogre::SHADOWTYPE_NONE);` is applied to disable shadows. However this is not the correct way to do it. This function disables shadow globally, [destroying](https://github.com/OGRECave/ogre/blob/master/OgreMain/src/OgreShadowRenderer.cpp#L1375-L1379) shadow maps (at least until the next frame, even if shadows are restored with this function), and is not meant as a way to quickly toggle on/off shadows. The [documentation of Ogre](https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_scene_manager.html#ad716d3752418408f073d9be9bbfe62d0) states:
> you should call this method at the start of the scene setup. 

This is the cause of issue https://github.com/osrf/gazebo/issues/2584. Basically, any camera that tries to render after a depth camera in the same frame won't have any shadow.

This is fixed this by removing the calls to `setShadowTechnique`. Instead, shadows are disabled on the viewport object. I did observe that this fixes the issue mentioned above in the project I'm working on.


iche033:

iche033:
Triggered another round CI. Local testing confirmed that the changes fix the issue for me.

iche033 (approved):

scpeters:
thanks for the fix!

PR #2914, opened 2021-01-12
Generate spot light shadow maps

iche033:
Updated gazebo to generate shadow maps for spot lights. This does not mean the spot lights will now start casting shadows in gazebo - we still need to update the RTShaderSystem code to actually render them. This does however mean that the shadow maps are now available and we can access them through custom material shaders.

I added a `spotlight_shadow_demo.world` that contains a shadow receiver ground plane with custom material shaders to show how to access the shadow texture generated by OGRE.

spot_light_shadow_receiver


gazebo/rendering/RTShaderSystemPrivate.hh
```diff
@@ -56,6 +56,9 @@ namespace gazebo
       /// \brief Flag to indicate that shaders need to be updated.
       public: bool updateShaders;
 
+      /// \brief Flag to indicate that shaders need to be updated.
```
scpeters: "shaders" -> "shadows"

gazebo/rendering/RTShaderSystem.cc
```diff
@@ -717,3 +685,91 @@ double RTShaderSystem::ShadowSplitPadding() const
 {
   return this->dataPtr->shadowSplitPadding;
 }
+
+/////////////////////////////////////////////////
+void RTShaderSystem::UpdateShadows()
+{
+  this->dataPtr->updateShadows = true;
```
scpeters, line 701: do we need to worry about tthread-safety here?

gazebo/rendering/RTShaderSystem.cc
```diff
@@ -717,3 +685,91 @@ double RTShaderSystem::ShadowSplitPadding() const
...
+#if defined(__APPLE__)
+  // workaround a weird but on OSX if texture size at 2 and 3 splits are not
+  // halved
+  texSize = this->dataPtr->shadowTextureSize/2;
+#endif
+  sceneMgr->setShadowTextureConfig(0,
+      this->dataPtr->shadowTextureSize, this->dataPtr->shadowTextureSize,
+      Ogre::PF_FLOAT32_R, 1);
+  sceneMgr->setShadowTextureConfig(1, texSize, texSize, Ogre::PF_FLOAT32_R, 2);
+  sceneMgr->setShadowTextureConfig(2, texSize, texSize, Ogre::PF_FLOAT32_R, 3);
```
scpeters: I believe the last arguments have been added since this code was moved down here (`fsaa` parameter with values 1, 2, 3); just confirming that this is intentional

gazebo/rendering/Light.cc
```diff
@@ -584,13 +585,19 @@ void Light::SetRange(const double _range)
...
+  this->dataPtr->light->setCastShadows(_cast);
+
+  if (_cast && this->dataPtr->light->getType() !=
+      Ogre::Light::LT_DIRECTIONAL)
   {
-    this->dataPtr->light->setCastShadows(false);
+    if (this->dataPtr->shadowCameraSetup.isNull())
+    {
+      auto *setup = new Ogre::FocusedShadowCameraSetup();
+      this->dataPtr->shadowCameraSetup = Ogre::ShadowCameraSetupPtr(setup);
```
scpeters: nit: consider using `std::make_shared` to combine these two lines if the type conversions permit it

scpeters (commented):
I haven't tested it  yet, but I made a few small comments and noticed from CI some failures in RenderingLight_TEST

* https://build.osrfoundation.org/job/gazebo-ci-pr_any-ubuntu_auto-amd64-gpu-nvidia/532/testReport/junit/(root)/Light_TEST/CastShadows/
* https://build.osrfoundation.org/job/gazebo-ci-pr_any-homebrew-amd64/2290/testReport/junit/(root)/Light_TEST/CastShadows/

gazebo/rendering/Light.cc
```diff
@@ -584,13 +585,19 @@ void Light::SetRange(const double _range)
 //////////////////////////////////////////////////
 void Light::SetCastShadows(const bool _cast)
 {
-  if (this->dataPtr->light->getType() == Ogre::Light::LT_DIRECTIONAL)
-  {
-    this->dataPtr->light->setCastShadows(_cast);
-  }
-  else
+  this->dataPtr->light->setCastShadows(_cast);
```
scpeters: should we call this for point lights?

scpeters (commented):
> noticed from CI some failures in RenderingLight_TEST

I believe [this failing test expectation for spot lights](https://github.com/osrf/gazebo/blob/spotlight_shadows_test/gazebo/rendering/RenderingLight_TEST.cc#L118-L119) should be reversed, but there's also a [failing expectation for point lights](https://github.com/osrf/gazebo/blob/spotlight_shadows_test/gazebo/rendering/RenderingLight_TEST.cc#L129-L130) and I'm not sure what that value should be, if shadows aren't yet working for point lights.

gazebo/rendering/Light.cc
iche033: I think the code should make point light generate shadow maps but because I didn't write any shaders and demo world to verify this, I decided to disable point light shadows for now. 7aa1d35.

gazebo/rendering/Light.cc
iche033: `Ogre::ShadowCameraSetupPtr` is an ogre shared pointer and not a `std::shared_pointer` so I was not able to use `make_shared`. I've combined the two lines nevertheless. 7aa1d35.

gazebo/rendering/RTShaderSystem.cc
iche033: thanks for catching this! that was for testing. I removed them. 7aa1d35.

gazebo/rendering/RTShaderSystem.cc
iche033: added mutex to protect the variable. 7aa1d35.

gazebo/rendering/RTShaderSystemPrivate.hh
iche033: fixed. 7aa1d35.

iche033:
>  haven't tested it yet, but I made a few small comments and noticed from CI some failures in RenderingLight_TEST

should be fixed. 7aa1d35.

scpeters:
> > haven't tested it yet, but I made a few small comments and noticed from CI some failures in RenderingLight_TEST
> 
> should be fixed. [7aa1d35](https://github.com/osrf/gazebo/commit/7aa1d3521b6b5da5f411011b2960656d0ecd2f30).

yep, the test is passing now :+1:

scpeters:
the spotlight shadows are working on Ubuntu, though they don't seem to be working on macOS. I do see a light grey artifact that changes orientation as I rotate the camera, but not the nice shadows generated on Ubuntu

I tried adding an extra spotlight just for fun in Ubuntu, and I didn't notice an extra dark area where it was shaded by both spot lights. the shadow darkness seems to be uniform color; maybe that is expected, I don't know

Screenshot from 2021-01-14 16-19-07

gazebo/rendering/RTShaderSystem.cc
```diff
@@ -636,20 +593,33 @@ Ogre::PSSMShadowCameraSetup *RTShaderSystem::GetPSSMShadowCameraSetup() const
...
 {
   GZ_PROFILE("rendering::RTShaderSystem::Update");
-  if (!this->dataPtr->initialized || !this->dataPtr->updateShaders)
+  if (!this->dataPtr->initialized)
   {
     return;
   }
 
-  for (const auto &scene : this->dataPtr->scenes)
+  std::lock_guard<std::mutex> lock(this->dataPtr->updateMutex);
```
scpeters: nit: if we want to minimize the time spent in this lock, we could only lock it while copying the values of `this->dataPtr->updateShadows` and `this->dataPtr->updateShaders`, then reset both to `false`, and then release the mutex

for example:

~~~
bool updateShaders, updateShadows;
{
  std::lock_guard<std::mutex> lock(this->dataPtr->updateMutex);
  updateShaders = this->dataPtr->updateShaders;
  updateShadows = this->dataPtr->updateShadows;
  this->dataPtr->updateShaders = false;
  this->dataPtr->updateShadows = false;
}
~~~

then read from the local variable instead of the member variable

iche033: yep done. 5f426b2

iche033:
> the spotlight shadows are working on Ubuntu, though they don't seem to be working on macOS. I do see a light grey artifact that changes orientation as I rotate the camera, but not the nice shadows generated on Ubuntu

I am guessing it could be related to the glsl version used. I remember we had issue with glsl 130 on macOS so I downgraded to version 120 in cf404b1. Let me know if this makes any difference.

> I tried adding an extra spotlight just for fun in Ubuntu, and I didn't notice an extra dark area where it was shaded by both spot lights. the shadow darkness seems to be uniform color; maybe that is expected, I don't know

yes that's expected. I wrote a very simple shader for the demo that just sets the color to a fixed value if it's a shadowed region. It's mainly for demo purposes and  the goal is to show that you can access the shadow map texture.

scpeters:
looks good to me; I'll approve once the CI finishes assuming there are no related regressions

scpeters (approved):

PR #2818, opened 2020-08-08
DARTHeightmapShape: correctly load asymmetric terrains

scpeters:
Support for heightmaps in DART was added in [bitbucket pull request 2956](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/2956) and [bitbucket pull request 3066](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/3066) with tests using the [heightmap_valley.png](https://github.com/osrf/gazebo/blob/gazebo11/media/materials/textures/heightmap_valley.png) image, which happens to be symmetric in both x and y directions. I recently noticed a bug in DARTHeightmapShape when using asymmetric terrains, so I added a test using [worlds/heightmap.world](https://github.com/osrf/gazebo/blob/gazebo11/worlds/heightmap.world), which uses the asymmetric [heightmap_bowl.png](https://github.com/osrf/gazebo/blob/gazebo11/media/materials/textures/heightmap_bowl.png) with a box at each of the 4 corners. The boxes all start at the same height, and based on the shape of the heightmap, 3 of the boxes should already be in contact with the heightmap, but one of them (`box2`) should fall. I've added a test that illlustrates the failure with dart:

~~~
[Dbg] [DARTPhysics.cc:734] Using BULLET collision detector
Dbg DART Bullet heightfield AABB: min = {-64.249, -64.249, -5}, max = {64.249, 64.249, 5} (will be translated by z=5)
AL lib: (EE) ALCpulsePlayback_streamStateCallback: Received stream failure!
[Dbg] [ServerFixture.cc:203] ServerFixture load in 0.8 seconds, timeout after 600 seconds
/data_fast/scpeters/ws/gazebo8/src/gazebo/test/integration/heightmap.cc:952: Failure
Expected: (box1->WorldPose().Pos().Z()) >= (9.9), actual: 5.0951 vs 9.9
/data_fast/scpeters/ws/gazebo8/src/gazebo/test/integration/heightmap.cc:956: Failure
Expected: (box2->WorldPose().Pos().Z()) < (5.5), actual: 9.98575 vs 5.5
[Dbg] [ServerFixture.cc:129] ServerFixture::Unload
~~~

The fix is to set an uninitialized variable `DARTHeightmapShape::flipY` to false.

Use `gazebo --verbose worlds/heightmap.world -e dart` to see the failure in gzclient.

PR #2781, opened 2020-07-09
Set friction params in DARTCollision for dart 6.10

scpeters:
They are currently being set in DARTLink; set them in DARTCollision instead if dart 6.10 is in use. Also for 6.10, set the friction direction parameter for DART and enable corresponding tests.

PR #2923, opened 2021-01-24
Merge 10 -> 11

scpeters:
Merge forward #2781 and #2818 from `gazebo10`. The #2809 commit is listed also, but it was already merged forward in #2816, so there are no changes in the diff.


scpeters:
CI looks ok to me, merging forward

PR #2927, opened 2021-01-26
Lens flare cleanup and colorization

mogumbo:


media/materials/programs/camera_lens_flare_fs.glsl
```diff
@@ -30,11 +20,8 @@ vec3 lensflare(vec2 uv,vec2 pos)
 
   float ang = atan(main.y, main.x);
   float dist = length(main); dist = pow(dist,.1);
-  float n = noise(vec2((ang-time/9.0)*16.0,dist*32.0));
 
-  float f0 = 1.0/(length(uv-pos)*16.0/scale+1.0);
-
-  f0 = f0+f0*(sin((ang+time/18.0 + noise(abs(ang)+n/2.0)*2.0)*12.0)*.1+dist*.1+.8);
+  float f0 = 2.0/(length(uv-pos)*16.0/scale+1.0);
```
mogumbo: Changed 1.0 to 2.0 here. After removing the legacy animated effect (which was causing ugly visual artifacts) and cc() (which had almost no effect on color), I boosted this number to restore the LensFlare's original appearance.

In my opinion, we should not do this because it makes the LensFlare look washed out and maintains a very big hotspot over the light source. Those features would never have been there if the broken animated effect had never been included.

gazebo/rendering/LensFlare.hh
```diff
@@ -58,6 +58,10 @@ namespace gazebo
       /// \param[in] _scale Scale of lens flare
       public: void SetScale(const double _scale);
 
+      /// \brief Set the color of lens flare.
+      /// \param[in] _color Color of lens flare
+      public: void SetColor(const ignition::math::Vector3d _color);
```
scpeters: nit: pass arg by reference `const ignition::math::Vector3d &_color`

plugins/LensFlareSensorPlugin.hh
```diff
@@ -41,6 +42,14 @@ namespace gazebo
     public: virtual void Load(sensors::SensorPtr _sensor,
         sdf::ElementPtr _sdf);
 
+    /// \brief Set the scale of lens flare.
+    /// \param[in] _scale Scale of lens flare
+    public: void SetScale(double _scale);
+
+    /// \brief Set the color of lens flare.
+    /// \param[in] _color Color of lens flare
+    public: void SetColor(ignition::math::Vector3d _color);
```
scpeters: pass const reference here too

mogumbo: fixed

gazebo/rendering/LensFlare.hh
mogumbo: fixed

media/materials/scripts/gazebo.material
```diff
@@ -1571,19 +1570,15 @@ material Gazebo/CameraLensFlare
...
 
       texture_unit RT
       {
         tex_coord_set 0
         tex_address_mode border
         filtering linear linear linear
       }
-      texture_unit noiseRGBA
-      {
-        texture noise_rgba.png
```
iche033, line 1585: we can remove this png file as I don't think it's used else where:
https://github.com/osrf/gazebo/blob/9fba67cafb9838752c29f7a91a12c5e18e9f3dd7/media/materials/textures/noise_rgba.png

gazebo/rendering/LensFlare.cc
```diff
@@ -71,6 +70,13 @@ namespace gazebo
         this->scale = _scale;
       }
 
+      /// \brief Set the color of lens flare.
+      /// \param[in] _color Color of lens flare
+      public: void SetColor(const ignition::math::Vector3d _color)
```
iche033: pass by const ref here as well. For consistency with gazebo coding style, can you put `&` next to variable name, i.e. `&_color`?

media/materials/scripts/lens_flare.compositor
```diff
@@ -3,7 +3,7 @@ compositor CameraLensFlare/Default
   technique
   {
     // Temporary textures
-    texture rt0 target_width target_height PF_A8R8G8B8
+    texture rt0 target_width target_height PF_FLOAT32_RGB
```
iche033: cc @scpeters - this is somewhat related to issue #2928. We could just change the internal texture format directly if we don't want to have a new sdformat param to configure it.

mogumbo: Both issues would benefit from an sdf param because most cases can work with a smaller, integer texture. Always using a FLOAT32 texture should work for every case but often waste graphics memory.

media/materials/scripts/gazebo.material
mogumbo: done

gazebo/rendering/LensFlare.cc
mogumbo: done

media/materials/programs/camera_lens_flare_fs.glsl
scpeters: so do you recommend `1.0` or `2.0` here?

media/materials/scripts/lens_flare.compositor
scpeters: I'm working on a custom SDFormat parameter in https://github.com/scpeters/gazebo/commit/8e88380b5d577894c605ce713593cff82185c0b8, just need to finish writing the test

media/materials/programs/camera_lens_flare_fs.glsl
mogumbo: I prefer 1.0 because the legacy animation effect made the whole lens flare look overdone. I think a more realistic lens flare effect would have excluded it. The only reason to keep 2.0 is so that appearance does not change significantly.

media/materials/scripts/lens_flare.compositor
scpeters: I'm ready to approve all parts of this pull request except for this line. Can we revert it and come back to it later, perhaps in the context of #2928?

mogumbo: Agreed. Reverted to PF_A8R8G8B8

scpeters (approved):

plugins/LensFlareSensorPlugin.hh
```diff
@@ -28,7 +28,10 @@ namespace gazebo
   /// \brief Plugin that adds lens flare effect to a camera or multicamera
   /// sensor
   /// The plugin has the following optional parameter:
-  /// <scale>     Scale of lens flare. Must be greater than 0
+  /// <scale>       Scale of lens flare. Must be greater than 0
+  /// <color>       Color of lens flare.
+  /// \todo A potentially useful feature would be an option for contantly
```
iche033: `contantly` -> `constantly?`

iche033 (approved):
looks good to me.

plugins/LensFlareSensorPlugin.hh
mogumbo: fixed

scpeters:
cherry-picking forward to gazebo11 in https://github.com/osrf/gazebo/pull/2954

PR #2937, opened 2021-02-26
fix(GpuRaySensor): Validate scene existence

reinzor:


iche033:
the changes look fine. A null scene also suggests something went wrong with initializing the rendering engine so pointing to a bigger problem.

Can you sign your commit?

chapulina (approved):

scpeters:
thanks for the fix!

PR #2945, opened 2021-03-06
Optimize ODE collision detection for models with no collisions

iche033:
It was found that removing all the collisions in a world with large number of models and links reduced performance / RTF. The problem is in ODE's collision checker. Models without collisions have a zero size bounding box. The collision checker seems to treat them all as overlapping collisions and hence results in extra unneeded computations. The "fix" is to skip collision checking for ODE geoms that have zero size AABB.

To reproduce this issue, you can run the following two worlds below. In both cases, go to Physics on the left panel and set `real_time_update_rate` to 0 for gazebo to run unbounded:

[dp.world](https://gist.github.com/iche033/02ad57d866b4e6783c83fa26d3b44f02): 100 static double pendulums with collisions. **RTF: 24.0**
[dp_no_collisions.world](https://gist.github.com/iche033/931efa8678ea8e9ba3ff2f2a638a7b4e): 100 static double pendulums with no collisions. Before fix: **RTF: 1.5**. After fix: **28.0**

Here're timings from the Remotery profiler 

dp.world. `dSpaceCollide`: 0.034ms
dp

dp_no_collisions.world - Before fix. `dSpaceCollide`: 0.635
dp_no_collision_before

dp_no_collisions.world - After fix. `dSpaceCollide`: 0.007ms
dp_no_collision_after

### DIFF
--- a/deps/opende/src/collision_space.cpp
+++ b/deps/opende/src/collision_space.cpp
@@ -452,6 +452,13 @@ void dxHashSpace::collide (void *_data, dNearCallback *callback)
     if (!GEOM_ENABLED(geom)){
       continue;
     }
+
+    // skip geoms that have zero or negative/invalid aabb
+    if (geom->aabb[1] - geom->aabb[0] <= 0.0 &&
+        geom->aabb[3] - geom->aabb[2] <= 0.0 &&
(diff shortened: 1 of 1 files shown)


scpeters (approved):
makes sense to me; thanks!

PR #2953, opened 2021-03-30
fixed marker

Mario-projects-dev:
fixed marker not behaving as expected due to id 0


scpeters:
this doesn't compile. I'm going to decline for now. perhaps you can create an issue to describe the problem you are trying to fix?

PR #2960, opened 2021-04-07
Specify wide angle camera cube map texture format

### LINKED ISSUE #2928: WideAngleCamera cube map texture pixel format should be customizable
The `rendering::WideAngleCamera::CreateEnvRenderTexture` function creates an `envCubeMapTexture` using the [same pixel format](https://github.com/osrf/gazebo/blob/gazebo11_11.3.0/gazebo/rendering/WideAngleCamera.cc#L663) as the output image. For advanced users who write their own shaders, it can be necessary to use a floating point pixel format for the cube map texture even if the output image uses integers, for example to avoid integer overflow. I think it would be convenient to specify this in SDFormat as a `env_texture_format` parameter that is a sibling to the [//camera/lens/env_texture_size](https://github.com/osrf/sdformat/blob/sdformat10_10.2.0/sdf/1.6/camera.sdf#L133-L135) element.

audrow:
This PR sets the texture format based on the custom element `//camera/lens/gazebo:env_texture_format`.

Fixes #2928.


gazebo/rendering/WideAngleCamera.cc
```diff
@@ -467,6 +471,13 @@ void WideAngleCamera::Load()
 
     if (sdfLens->HasElement("env_texture_size"))
       this->dataPtr->envTextureSize = sdfLens->Get<int>("env_texture_size");
+
+    const std::string envTextureFormat = "gazebo:env_texture_format";
```
scpeters: cc @chapulina @iche033 should we change this xml namespace to `ignition:` to match the Mecanum wheel parameter namespace?

test/integration/wideanglecamera_sensor.cc
```diff
@@ -225,3 +225,112 @@ TEST_F(WideAngleCameraSensor, Projection)
...
+      unsigned int g = img[i+1];
+      unsigned int b = img[i+2];
+      rSum += r;
+      gSum += g;
+      bSum += b;
+    }
+
+    // For grayscale, all RGB channels should have the same value
+    EXPECT_DOUBLE_EQ(rSum, gSum);
+    EXPECT_DOUBLE_EQ(gSum, bSum);
```
scpeters: nit: replace `EXPECT_DOUBLE_EQ` with `EXPECT_EQ` on these two lines, since it is comparing integers

test/integration/wideanglecamera_sensor.cc
```diff
@@ -225,3 +225,112 @@ TEST_F(WideAngleCameraSensor, Projection)
...
+    EXPECT_NE(gSum, 0.0);  // check that image isn't black
```
scpeters: I would strengthen this expectation to the following:

~~~
EXPECT_GT(gSum, 10000u);
~~~

gazebo/rendering/WideAngleCamera.cc
scpeters: I asked them on slack, and they recommend changing the xml namespace prefix from `gazebo` to `ignition`

we also need to declare the xml namespace; I'll add an inline comment on the world file

worlds/test/issue_2928_wide_angle_camera_texture_format.world
```diff
@@ -0,0 +1,140 @@
...
+            <clip>
+              <near>0.1</near>
+              <far>100</far>
+            </clip>
+            <lens>
+              <type>gnomonical</type>
+              <scale_to_hfov>1</scale_to_hfov>
+              <cutoff_angle>1.5707</cutoff_angle>
+              <env_texture_size>512</env_texture_size>
+              <gazebo:env_texture_format>R_FLOAT16</gazebo:env_texture_format>
```
scpeters: we should declare an xml namespace by adding the `xmlns` attribute to a containing element (either the model or the world). For more context, see the [SDFormat documentation about custom elements](http://sdformat.org/tutorials?tut=custom_elements_attributes_proposal&cat=pose_semantics_docs&#specifying-custom-elements-and-attributes) and an [example from ign-gazebo](https://github.com/ignitionrobotics/ign-gazebo/blob/26832764749291b6ea22b525262136ecbb63fbb9/examples/worlds/mecanum_drive.sdf#L80-L158)

per the comment thread above, I recommend using `xmlns:ignition="http://ignitionrobotics.org/schema"` and `<ignition:env_texture_format/>`

scpeters (approved):

scpeters:
CI looks ok to me

iche033 (commented):
looks good to me.

iche033 (approved):

PR #2965, opened 2021-04-12
LensFlare - allow inheritance 

kjeppesen1:
Added LensFlarePrivate setter methods and moved LensFlareCompositor definition to header

### DIFF
--- a/gazebo/rendering/LensFlare.cc
+++ b/gazebo/rendering/LensFlare.cc
@@ -33,7 +33,6 @@ namespace gazebo
 {
   namespace rendering
   {
-
     /// \brief Private data class for LensFlareCompositorListener
     class LensFlareCompositorListenerPrivate
     {
@@ -58,7 +57,8 @@ namespace gazebo
     };
 
     //////////////////////////////////////////////////
-    LensFlareCompositorListener::LensFlareCompositorListener(CameraPtr _camera, LightPtr _light)
+    LensFlareCompositorListener::LensFlareCompositorListener(CameraPtr _camera,
+                                                             LightPtr _light)
       : dataPtr(new LensFlareCompositorListenerPrivate)
     {
       this->dataPtr->camera = _camera;
@@ -88,7 +88,8 @@ namespace gazebo
     }
 
     //////////////////////////////////////////////////
-    void LensFlareCompositorListener::SetColor(const ignition::math::Vector3d &_color)
+    void LensFlareCompositorListener::SetColor(
+        const ignition::math::Vector3d &_color)
     {
       this->dataPtr->color = _color;
     }
@@ -118,8 +119,10 @@ namespace gazebo
 
       // for adjusting aspect ratio of flare
       params->setNamedConstant("viewport",
-          Ogre::Vector3(static_cast<double>(this->dataPtr->camera->ViewportWidth()),
-          static_cast<double>(this->dataPtr->camera->ViewportHeight()), 1.0));
+          Ogre::Vector3(
+              static_cast<double>(this->dataPtr->camera->ViewportWidth()),
+              static_cast<double>(this->dataPtr->camera->ViewportHeight()),
+              1.0));
 
       // use light's world position for lens flare position
       if (this->dataPtr->light->Type() == "directional")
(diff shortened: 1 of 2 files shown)


kjeppesen1:
@scpeters @iche033

gazebo/rendering/LensFlare.cc
```diff
@@ -392,6 +77,280 @@ namespace gazebo
 using namespace gazebo;
 using namespace rendering;
 
+//////////////////////////////////////////////////
+LensFlareCompositorListener::LensFlareCompositorListener(CameraPtr _camera, LightPtr _light)
```
scpeters: I think it would simplify the diff if we moved the implementation of the `LensFlareCompositorListener` methods back so they are above `LensFlarePrivate` again. The whitespace is negotiable, since we can ignore it with `?w=1` added to the URL

gazebo/rendering/LensFlare.hh
```diff
@@ -31,7 +32,80 @@ namespace gazebo
...
+
+      /// \brief Position of light in world frame
+      private: ignition::math::Vector3d lightWorldPos;
+
+      /// \brief Scale of lens flare.
+      private: double scale = 1.0;
+
+      /// \brief Color of lens flare.
+      private: ignition::math::Vector3d color
+          = ignition::math::Vector3d(1.0, 1.0, 1.0);
```
scpeters: if we are adding this class to the public API, then it should use the PIMPL pattern to allow more flexibility without breaking ABI

scpeters (commented):
the ABI checker job failed; I'll see if I can get it running again

gazebo/rendering/LensFlare.hh
kjeppesen1: Just to be sure that I am understanding your request correctly, you are suggesting to replace all the private member variables (i.e. `light` through `color`) with a `private: std::unique_ptr<LensFlareCompositorListenerPrivate> dataPtr`, where `LensFlareCompositorListenerPrivate` will be defined in the source file, just like how it is done with `LensFlare` and `LensFlarePrivate`?

scpeters: yes that is what I am proposing

I can assist with the modifications if needed

scpeters: thanks, on first look the PIMPL implementation looks great!

gazebo/rendering/LensFlare.cc
scpeters: thanks, the `?w=1` diff is much cleaner now!

scpeters:
diff without whitespace:

* https://github.com/osrf/gazebo/pull/2965/files?w=1

gazebo/rendering/LensFlare.hh
kjeppesen1: FYI I'm still running some tests on my own end to verify it, just needed to push to test in a separate repo. One new addition is I need to init a LensFlareCompositorListenerPrivate in the constructor. I'll let you know when all my tests are working.

kjeppesen1: New commit fixes it and passes my tests :)

scpeters:
our jenkins builds are not properly reporting style errors (https://github.com/ignition-tooling/release-tools/issues/446), so I took the liberty of shortening some line lengths to 80 characters and cleaning up some whitespace in 48837d1

scpeters (approved):

gazebo/rendering/LensFlare.hh
```diff
@@ -63,7 +124,20 @@ namespace gazebo
       public: void SetColor(const ignition::math::Vector3d &_color);
 
       /// \brief Update function to search light source
-      private: void Update();
+      protected: void Update();
+
+      /// \brief Set camera stored in LensFlarePrivate
+      /// \param[in] _camera Camera to use in sensor.
+      protected: void SetCameraSensor(CameraPtr _camera);
```
iche033, line 131: note that `_camera` can be either a camera sensor or a user camera but this name suggests that it needs to be a camera sensor.

Maybe something like `SetCameraImp`?  We've used the the *Imp suffix in other places for internal functions.

iche033 (commented):
A little late. Changes look good. Just a minor comment about function name

gazebo/rendering/LensFlare.hh
scpeters: nice catch! updating to `LensFlare::SetCameraImpl` in #2975

PR #2966, opened 2021-04-13
Added reflectance output image for DepthCamera

WilliamLewww:
When specified in the SDF file, the depth camera will output a reflectance image that can be viewed in the topic visualizer.

reflectance


gazebo/rendering/DepthCamera.hh
```diff
@@ -92,6 +92,14 @@ namespace gazebo
       /// \return The z-buffer as a float array
       public: virtual const float *DepthData() const;
 
+      /// \brief Get whether reflectance has been generated.
+      /// \return True if reflectance is set to generate.
+      public: bool GetOutputReflectance() const;
+
+      /// \brief All things needed to get reflectance data
+      /// \return The reflectance buffer as a float array
+      public: virtual const float *ReflectanceData() const;
```
scpeters: marking this `virtual` breaks ABI

* [Build Status](https://build.osrfoundation.org/job/gazebo-abichecker-any_to_any-ubuntu_auto-amd64/804/) https://build.osrfoundation.org/job/gazebo-abichecker-any_to_any-ubuntu_auto-amd64/804/API_5fABI_20report/

can the `virtual` specifier be removed or is it needed?

WilliamLewww: I added the virtual to match the method above, `public: virtual const float *DepthData() const;`, but it is not necessary. I removed the virtual modifier and pushed an update.

gazebo/sensors/DepthCameraSensor.cc
```diff
@@ -188,6 +202,40 @@ bool DepthCameraSensor::UpdateImpl(const bool /*_force*/)
...
+      this->dataPtr->imageReflectancePub->HasConnections() &&
+      // check if reflectance data is available.
+      this->dataPtr->depthCamera->ReflectanceData())
+  {
+    msgs::ImageStamped msg;
+    msgs::Set(msg.mutable_time(), this->scene->SimTime());
+    msg.mutable_image()->set_width(this->camera->ImageWidth());
+    msg.mutable_image()->set_height(this->camera->ImageHeight());
+    msg.mutable_image()->set_pixel_format(common::Image::R_FLOAT32);
+
```
ahcorde:
```suggestion
```

ahcorde (changes requested):
can we add a test to check the ign topic ?

WilliamLewww:
I added a test for both enabled and disabled reflectance output.

gazebo/sensors/DepthCameraSensor_TEST.cc
```diff
@@ -193,6 +193,75 @@ TEST_F(DepthCameraReflectanceSensor_TEST, CreateDepthCamera)
...
+
+/////////////////////////////////////////////////
+/// \brief Callback executed for every propagation grid message received
+void TxMsg(const ConstPropagationGridPtr &_msg)
+{
+  std::lock_guard<std::mutex> lock(g_reflectanceTopicMutex);
+  // Just copy the message
+  g_reflectanceTopicGridMsg = _msg;
+  g_reflectanceTopicReceivedMsg = true;
+}
```
ahcorde: You should include this variables and function inside the class `DepthCameraSensor_reflectance_topic_TEST`

ahcorde (changes requested):

WilliamLewww:
Should I create a different pull request to encapsulate all the date for the other tests that are already there?

https://github.com/osrf/gazebo/blob/2e42eb91e7638da54b36dbb8fdd4ae5e3a310d6d/gazebo/sensors/DepthCameraSensor_TEST.cc#L25-L32

https://github.com/osrf/gazebo/blob/2e42eb91e7638da54b36dbb8fdd4ae5e3a310d6d/gazebo/sensors/DepthCameraSensor_TEST.cc#L196-L201

ahcorde:
> Should I create a different pull request to encapsulate all the date for the other tests that are already there?

You can do it in this PR

ahcorde (approved):
@steve do you mind to take another quick look ?

scpeters:
> Steve do you mind to take another quick look ?

I assume you meant me. Sure I'll take another look

gazebo/sensors/DepthCameraSensor.cc
```diff
@@ -124,6 +124,20 @@ void DepthCameraSensor::Init()
     gzerr << "No world name" << std::endl;
   }
 
+  if (this->dataPtr->depthCamera->GetOutputReflectance())
+  {
+    ignition::transport::AdvertiseMessageOptions opts;
+    opts.SetMsgsPerSec(50);
```
scpeters: should we use the `Sensor::UpdateRate()` instead of hard-coding to 50 Hz?

gazebo/sensors/DepthCameraSensor.cc
```diff
@@ -124,6 +124,20 @@ void DepthCameraSensor::Init()
...
+
+    this->dataPtr->imageReflectancePub =
+        this->node->Advertise<msgs::ImageStamped>(
+            this->Topic() + "_reflectance", 50);
+
+    this->dataPtr->imageReflectancePubIgn =
```
scpeters: I think this ignition topic is never published

gazebo/sensors/DepthCameraSensor.cc
WilliamLewww: I noticed that the other cameras are using a hard-coded 50Hz:

https://github.com/osrf/gazebo/blob/26b69a7cff29312164558d23344eb551a96508ba/gazebo/sensors/CameraSensor.cc#L110-L112

https://github.com/osrf/gazebo/blob/26b69a7cff29312164558d23344eb551a96508ba/gazebo/sensors/WideAngleCameraSensor.cc#L146-L148

When I tried changing the `50` to `this->UpdateRate()`, I get a broken topic because I think the update rate has not been set and still is the value 0.000. 

I could change this and initialize the update rate before advertising the topic but I am unsure if this will cause problems to the other components on the depth camera.

scpeters: `gazebo::transport::Node::Advertise` has [two optional numeric parameters](https://github.com/osrf/gazebo/blob/gazebo11_11.4.0/gazebo/transport/Node.hh#L176-L179), first the queue size and then the rate of publishing in Hz. In these examples from CameraSensor and WideAngleCameraSensor, `50` is actually the queue size. In the ignition::transport code (that has now been removed) it was a bit more clear that `SetMsgsPerSec` was about the rate of publishing

I think no additional changes are needed to the queue size

gazebo/sensors/DepthCameraSensor.cc
scpeters: I guess if we don't need the ignition publisher, then it's fine to remove. @ahcorde what do you think?

ahcorde (approved):

WilliamLewww:
Inverted the reflectance buffer to correct the inverted values for depth images.

a

scpeters:
I've merged with `gazebo11` to fix the abichecker build

gazebo/rendering/DepthCamera.hh
```diff
@@ -92,6 +92,14 @@ namespace gazebo
       /// \return The z-buffer as a float array
       public: virtual const float *DepthData() const;
 
+      /// \brief Get whether reflectance has been generated.
+      /// \return True if reflectance is set to generate.
+      public: bool GetOutputReflectance() const;
+
+      /// \brief All things needed to get reflectance data
+      /// \return The reflectance buffer as a float array
```
scpeters, line 100: can you clarify what the values in this array mean physically? does 0 mean nonreflective and 1 mean reflective?

I'm asking because you inverted the data in c776173ca7ff3b115bb691bb40499e2ed1d78621

iche033: oh we should not invert the data on the sensors side. We should only update the visualization code. The gui visualization inverts the data when it detects that the underlying image has a FLOAT type:
https://github.com/osrf/gazebo/blob/gazebo11/gazebo/gui/viewers/ImageFrame.cc#L157-L167

The logic there was added for visualization of depth data, i.e. black pixels means far away while white means close to camera. However, now that the reflectance image output is also in FLOAT, we don't want the inversion to be applied to the reflectance image. I think there needs to be some way on the gui side to tell reflectance images from depth images so that the correct visualization is applied.

WilliamLewww:
To prevent the buffer from inverting, I set the reflectance message format to `UNKNOWN_PIXEL_FORMAT` so that the image frame would be able to detect that the message isn't a depth image. The pixel format `R_FLOAT32` is hardcoded to be recognized as a depth buffer, and to maintain the pixel format while being able to distinguish `depth vs. reflectance` would require pixel analysis.

A cleaner solution is to add a new data member to the `Image` message specifying what type of image the message contains. The change could remove the hardcoded `pixel format to image type` mapping but will make the Image message larger. I am not too sure how disruptive this change may be so I opted for the first solution.

chapulina:
What's the status of this PR?

PR #2954, opened 2021-03-30
Cherry-pick #2927 to gazebo11: Lens flare cleanup and colorization

scpeters:
This is a cherry-pick forward of #2927 from gazebo9 to gazebo11. Use "rebase and merge".

* Added color setting to LensFlare and LensFlareSensorPlugin.
Added methods to expose scale and color to derived plugins.
* Removed unused animated lens flare effect.
* Removed noise_rgba.png, which is no longer used.
* Changed a value in the lens flare shader back to its original 1.0.
This means there will be no compensation for the removal of the
legacy animation effect and the lens flares will look less intense.

PR #2950, opened 2021-03-18
Avoid deadlock in ConnectionManager::Stop

ivanpauno:
Similarly to https://github.com/osrf/gazebo/pull/2949, it fixes another condition where `gzserver` hangs when shutting down.

The main thread was interrepted by sigint and is waiting for `ConnectionManager::Stop()` to finish:

```
Thread 1 (Thread 0x7f8c8c45a3c0 (LWP 12947)):
#0  0x00007f8c9edc43bf in __GI___clock_nanosleep (clock_id=clock_id@entry=1, flags=flags@entry=0, req=req@entry=0x7ffe7834a060, rem=rem@entry=0x7ffe7834a070) at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
#1  0x00007f8c9f3dfadc in gazebo::common::Time::Sleep(gazebo::common::Time const&) (_time=...) at /home/ivanpauno/ros2_ws/other_ws/gazebo/gazebo/common/Time.cc:441
#2  0x00007f8c9f3dfc5c in gazebo::common::Time::MSleep(unsigned int) (_ms=_ms@entry=100) at /home/ivanpauno/ros2_ws/other_ws/gazebo/gazebo/common/Time.cc:459
#3  0x00007f8c9e456a2d in gazebo::transport::ConnectionManager::Stop() (this=<optimized out>) at /home/ivanpauno/ros2_ws/other_ws/gazebo/gazebo/transport/ConnectionManager.cc:248
#4  gazebo::transport::ConnectionManager::Stop() (this=0x7f8c9fb44e60 <SingletonT<gazebo::transport::ConnectionManager>::GetInstance()::t>) at /home/ivanpauno/ros2_ws/other_ws/gazebo/gazebo/transport/ConnectionManager.cc:242
#5  0x00007f8c9faeb2dd in std::function<void ()>::operator()() const (this=<optimized out>) at /usr/include/c++/9/bits/std_function.h:683
#6  gazebo::event::EventT<void ()>::Signal() (this=<optimized out>) at /home/ivanpauno/ros2_ws/other_ws/gazebo/gazebo/common/Event.hh:283
#7  gazebo::event::EventT<void ()>::operator()() (this=<optimized out>) at /home/ivanpauno/ros2_ws/other_ws/gazebo/gazebo/common/Event.hh:123
#8  gazebo::Server::SigInt(int) () at /home/ivanpauno/ros2_ws/other_ws/gazebo/gazebo/Server.cc:571
[...]

PR #2968, opened 2021-04-14
Backport to 9: fixes for gzserver shutdown and ogre log test

scpeters:
This backports the following pull requests to gazebo9:

* ~~#2960: Specify wide angle camera cube map texture format~~ redacted because it doesn't work with libsdformat6
* #2949: improve gzserver shutdown by protecting DepthCameraPlugin globals with a mutex
* #2950: improve gzserver shutdown by avoiding deadlock in ConnectionManager::Stop
* #2954: fix ogre_log test

Use rebase and merge.

### DIFF
--- a/gazebo/transport/ConnectionManager.cc
+++ b/gazebo/transport/ConnectionManager.cc
@@ -215,6 +215,9 @@ void ConnectionManager::Fini()
     return;
 
   this->Stop();
+  if (this->initialized)
+    while (this->stopped == false)
+      common::Time::MSleep(100);
 
   if (this->masterConn)
   {
@@ -244,10 +247,6 @@ void ConnectionManager::Fini()
 void ConnectionManager::Stop()
 {
   this->stop = true;
-  this->updateCondition.notify_all();
-  if (this->initialized)
-    while (this->stopped == false)
-      common::Time::MSleep(100);
 }
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


worlds/test/issue_2928_wide_angle_camera_texture_format.world
```diff
@@ -0,0 +1,140 @@
+<?xml version="1.0" ?>
+<sdf version='1.7'>
```
iche033: will this work in gazebo9? it's using sdformat6 which has sdf spec up to 1.6

scpeters: good point. I think we can downgrade that version. I'll test locally

scpeters: the test still fails even after downgrading this to `1.6` since the libsdformat6 parser ignores unrecognized elements like `<ignition:env_texture_format>` (some history in https://github.com/osrf/sdformat/issues/327#issuecomment-668325613)

so we will need a different solution on `gazebo9`

scpeters: I'm going to redact #2960 from this PR because it won't work with libsdformat6

scpeters: redacted

iche033 (approved):

PR #2963, opened 2021-04-09
Remove warning in message conversion using gpu_ray sensor

eborghi10:
This PR attempts to fix #2355 and get rid of an annoying message when the `verbose` flag is enabled.

PR #2972, opened 2021-04-16
Remove warning message in verbose mode when using GPU rays or depth cameras

eborghi10:
Backported PR #2963.

- In #2963, the fix for the depth camera was not introduced so another PR will be created later.

### Testing

```bash
gazebo --verbose test/worlds/gpu_laser2.world

gazebo --verbose worlds/depth_camera2.world
```

Path are relative to Gazebo.


scpeters (approved):

scpeters:
thanks for back-porting!

PR #2974, opened 2021-04-17
Cherry-pick LensFlare api changes to 11

scpeters:
Cherry-pick the following pull requests forward to gazebo11. Use rebase and merge.

* #2965 LensFlare: allow inheritance. This add to the public API of LensFlare
* #2975 LensFlare: rename SetCameraSensor to SetCameraImpl
* #2947 Fix size of spotlight visual
* #2907 Fix depth camera breaking shadows

PR #2975, opened 2021-04-19
LensFlare: rename SetCameraSensor to SetCameraImpl

scpeters:
This is an adjustment to API added in #2965 in response to https://github.com/osrf/gazebo/pull/2965#discussion_r615210697, which is allowed because the API has not yet been released.

### DIFF
--- a/gazebo/rendering/LensFlare.cc
+++ b/gazebo/rendering/LensFlare.cc
@@ -525,7 +525,7 @@ void LensFlare::OnRequest(ConstRequestPtr &_msg)
 }
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


scpeters:
> This is an adjustment to API ... which is allowed because the API has not yet been released.

the following is expected and not a problem

[Build Status](https://build.osrfoundation.org/job/gazebo-abichecker-any_to_any-ubuntu_auto-amd64/828/) https://build.osrfoundation.org/job/gazebo-abichecker-any_to_any-ubuntu_auto-amd64/828/API_5fABI_20report/

scpeters:
merging without review so I can cherry-pick it into #2974

PR #2982, opened 2021-04-22
Fix light preview parameters

### LINKED ISSUE #2981: Temporary GUI light has different angle than when it is added to the scene
With #2947 merged, it is now even more apparent that the temporary spot light used in GUI when interactively adding lights has wrong "shining angle":

https://user-images.githubusercontent.com/182533/115684549-bdfeb280-a357-11eb-880f-90fbee9006a2.mp4

https://user-images.githubusercontent.com/182533/115684562-bfc87600-a357-11eb-96fa-c6f096f0b351.mp4

As can be seen in the videos, the wrong size now affects both the light source and the visual (i.e. the visual's size is correct regarding the temporary light's size).

Or is it just so that the temporary light has different default properties than the default added light?

peci1:
Fixes #2981 . See the issue for detailed description.

https://user-images.githubusercontent.com/182533/115697965-661a7880-a364-11eb-9ac8-74f06fbdc843.mp4

I also wanted to add tests for the inner/outer angle, but they have no getters. Should I add these? Please, also check that the copy of `light` I make in the test is actually still a pointer to the preview light and is not pointing to the same object `light` does at the end of the test. I think it should be so, but another pair of eyes would be good :)

### DIFF
--- a/gazebo/rendering/Light.cc
+++ b/gazebo/rendering/Light.cc
@@ -156,14 +156,20 @@ void Light::UpdateSDFFromMsg(const msgs::Light &_msg)
 //////////////////////////////////////////////////
 void Light::UpdateFromMsg(ConstLightPtr &_msg)
 {
-  this->UpdateSDFFromMsg(*_msg);
+  this->UpdateFromMsg(*_msg);
+}
+
+//////////////////////////////////////////////////
+void Light::UpdateFromMsg(const msgs::Light &_msg)
+{
+  this->UpdateSDFFromMsg(_msg);
 
   this->Update();
 
-  if (_msg->has_pose())
+  if (_msg.has_pose())
   {
-    this->SetPosition(msgs::ConvertIgn(_msg->pose().position()));
-    this->SetRotation(msgs::ConvertIgn(_msg->pose().orientation()));
+    this->SetPosition(msgs::ConvertIgn(_msg.pose().position()));
+    this->SetRotation(msgs::ConvertIgn(_msg.pose().orientation()));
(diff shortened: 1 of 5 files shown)


ahcorde (approved):
Thank you for the fix, 

Regarding the inner/outer angles tests, Adding new methods don't break API, I think it's fine to add them, @chapulina or @scpeters Do you agree?

gazebo/rendering/Light.cc
```diff
@@ -156,14 +156,20 @@ void Light::UpdateSDFFromMsg(const msgs::Light &_msg)
 //////////////////////////////////////////////////
 void Light::UpdateFromMsg(ConstLightPtr &_msg)
 {
-  this->UpdateSDFFromMsg(*_msg);
+  this->UpdateFromMsg(*_msg);
```
scpeters: I think this `_msg` pointer should be checked before dereferencing it

scpeters (commented):
> Regarding the inner/outer angles tests, Adding new methods don't break API, I think it's fine to add them, @chapulina or @scpeters Do you agree?

you are welcome to add getters for inner and outer angles if you like

gazebo/rendering/Light.cc
peci1: Resolved (it ignores null messages).

peci1:
I added the getters and added a test for spotlight creation which utilizes the getters.

chapulina (approved):

chapulina:
New test passed on CI, merging!

Issue #2527, opened 2018-09-18
Last camera's distortion affects previous cameras' distortions


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/2527)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
See the following world for example, where the first camera appears to have barrel distortion even though it was supposed to be pincushion. Swapping the order of models in the world makes both cameras be pincushion.

~~~
<?xml version="1.0" ?>
<sdf version="1.6">
  <world name="default">

    <include>
      <uri>model://sun</uri>
    </include>

    <model name="ground_plane">
      <static>true</static>
      <link name="link">
        <collision name="collision">
          <geometry>
            <plane>
              <normal>0 0 1</normal>
              <size>100 100</size>
            </plane>
          </geometry>
          <surface>
            <friction>
              <ode>
                <mu>100</mu>
                <mu2>50</mu2>
              </ode>
            </friction>
              <bounce>
                <restitution_coefficient>1.0</restitution_coefficient>
                <threshold>0</threshold>
              </bounce>
              <contact>
                <ode>
                  <max_vel>10000</max_vel>
                  <min_depth>0.001</min_depth>
                </ode>
              </contact>
          </surface>
        </collision>
        <visual name="visual">
          <cast_shadows>false</cast_shadows>
          <geometry>
            <plane>
              <normal>0 0 1</normal>
              <size>100 100</size>
            </plane>
          </geometry>
          <material>
            <script>
              <uri>file://media/materials/scripts/gazebo.material</uri>
              <name>Gazebo/Grey</name>
            </script>
          </material>
        </visual>
      </link>
    </model>

    <model name="ball">
      <allow_auto_disable>false</allow_auto_disable>
      <pose>-1 0 1 0 0 0</pose>
      <static>false</static>
      <link name="link">
        <inertial>
          <mass>0.026</mass>
          <!-- inertia based on solid sphere 2/5 mr^2 -->
          <inertia>
            <ixx>1.664e-5</ixx>
            <iyy>1.664e-5</iyy>
            <izz>1.664e-5</izz>
            <ixy>0</ixy>
            <ixz>0</ixz>
            <iyz>0</iyz>
          </inertia>
[...]

PR #3009, opened 2021-05-22
Distortion normalization improvement + fix "folding" on large distortions

### LINKED ISSUE #3003: Distortion implementation correction
When undistorting a distorted image taken in Gazebo, the focal length must be specified as the image resolution for it to be correct.

I believe the issue is in the "normalization" of the image pixels that are put into Brown's Distortion Equations. I have attached some slides explaining the fix, including some tests performed to demonstrate it. The short description is that the distortion must occur in the camera plane rather than the image plane, and thus the pixel coordinates must first be projected to the camera plane via dividing by the focal length. Then after distortion the focal length is then multiplied to project the image back to the image plane.
[Gazebo Distortion Modification.pdf](https://github.com/osrf/gazebo/files/6468096/Gazebo.Distortion.Modification.pdf)

The relevant function is [Distortion::Distort()](https://github.com/osrf/gazebo/blob/e2436b3a2b32713e2b069f23689b62786273d773/gazebo/rendering/Distortion.cc#L385).

There is a minor change to the math performed at the distortion, and there is a need for Distortion.cc to have access to the focal length parameter (not implemented here). Here is a fix for the function itself that I tested with:

```
ignition::math::Vector2d Distortion::Distort(
    const ignition::math::Vector2d &_in,
    const ignition::math::Vector2d &_center, double _k1, double _k2, double _k3,
    double _p1, double _p2, unsigned int _width, double _f)
{
  // apply Brown's distortion model, see
  // http://en.wikipedia.org/wiki/Distortion_%28optics%29#Software_correction

  ignition::math::Vector2d normalized2d = ((_in*_width) - (_center*_width))/_f;
  ignition::math::Vector3d normalized(normalized2d.X(), normalized2d.Y(), 0);
  double rSq = normalized.X() * normalized.X() +
               normalized.Y() * normalized.Y();

  // radial
  ignition::math::Vector3d dist = normalized * (1.0 +
      _k1 * rSq +
      _k2 * rSq * rSq +
      _k3 * rSq * rSq * rSq);

  // tangential
  dist.X() += _p2 * (rSq + 2 * (normalized.X()*normalized.X())) +
      2 * _p1 * normalized.X() * normalized.Y();
  dist.Y() += _p1 * (rSq + 2 * (normalized.Y()*normalized.Y())) +
      2 * _p2 * normalized.X() * normalized.Y();

  return ((_center*_width) + ignition::math::Vector2d(dist.X(), dist.Y())*_f)/_width;
}
```

kjeppesen1:
Fixes #3003.

Modifies how Brown's distortion equations are applied to better reflect real distortion. Image is projected from image plane to camera plane to apply distortion equations, then projected back to image plane.

Before/after focal scaling implementation:
distortion_comparisons_gazebo

Before/after extreme distortion edge case fix:
before fix after fix

---

We use a tag `<ignition:legacy_mode>` in a distortion context to select this new behavior.

Regardless of if the new mode or legacy mode are used, this PR adds a bug fix for heavy distortions that prevents the distorted image from "folding back" on itself, as seen in the image above.

### DIFF
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -201,7 +201,7 @@ void Distortion::SetCamera(CameraPtr _camera)
       this->dataPtr->lensCenter.Y() * this->dataPtr->distortionTexWidth);
 
   // declare variables before the loop
-  static auto unsetPixelVector =  ignition::math::Vector2d(-1, -1);
+  const auto unsetPixelVector =  ignition::math::Vector2d(-1, -1);
   ignition::math::Vector2d normalizedLocation,
       distortedLocation,
       newDistortedCoordinates,
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -174,9 +174,9 @@ void Distortion::SetCamera(CameraPtr _camera)
   unsigned int texSide = _camera->ImageHeight() > _camera->ImageWidth() ?
       _camera->ImageHeight() : _camera->ImageWidth();
   // calculate focal length from largest fov
-  double fov = _camera->ImageHeight() > _camera->ImageWidth() ?
+  const double fov = _camera->ImageHeight() > _camera->ImageWidth() ?
       _camera->VFOV().Radian() : _camera->HFOV().Radian();
-  double focalLength = texSide/(2*tan(fov/2));
+  const double focalLength = texSide/(2*tan(fov/2));
   this->dataPtr->distortionTexWidth = texSide - 1;
   this->dataPtr->distortionTexHeight = texSide - 1;
   unsigned int imageSize =
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -212,10 +212,12 @@ void Distortion::SetCamera(CameraPtr _camera)
   double normalizedColLocation, normalizedRowLocation;
 
   // fill the distortion map
-  for (unsigned int mapRow = 0; mapRow < this->dataPtr->distortionTexHeight; ++mapRow)
+  for (unsigned int mapRow = 0; mapRow < this->dataPtr->distortionTexHeight;
+    ++mapRow)
   {
     normalizedRowLocation = mapRow*rowStepSize;
-    for (unsigned int mapCol = 0; mapCol < this->dataPtr->distortionTexWidth; ++mapCol)
+    for (unsigned int mapCol = 0; mapCol < this->dataPtr->distortionTexWidth;
+      ++mapCol)
     {
       normalizedColLocation = mapCol*colStepSize;
 
@@ -240,25 +242,29 @@ void Distortion::SetCamera(CameraPtr _camera)
 
       // compute the index in the distortion map
       distortedCol = distortedLocation.X() * this->dataPtr->distortionTexWidth;
-      distortedRow = distortedLocation.Y() * this->dataPtr->distortionTexHeight;
+      distortedRow = distortedLocation.Y() *
+        this->dataPtr->distortionTexHeight;
 
       // Note that the following makes sure that, for significant distortions,
-      // there is not a problem where the distorted image seems to fold over itself.
-      // This is accomplished by favoring pixels closer to the center of distortion,
-      // and this change applies to both the legacy and nonlegacy distortion modes.
+      // there is not a problem where the distorted image seems to fold over
+      // itself. This is accomplished by favoring pixels closer to the center
+      // of distortion, and this change applies to both the legacy and
+      // nonlegacy distortion modes.
 
       // Make sure the distorted pixel is within the texture dimensions
       if (distortedCol < this->dataPtr->distortionTexWidth &&
           distortedRow < this->dataPtr->distortionTexHeight)
       {
-        distortedIdx = distortedRow * this->dataPtr->distortionTexWidth + distortedCol;
+        distortedIdx = distortedRow * this->dataPtr->distortionTexWidth +
+          distortedCol;
 
         // check if the index has already been set
         if (this->dataPtr->distortionMap[distortedIdx] != unsetPixelVector)
         {
           // grab current coordinates that map to this destination
-          currDistortedCoordinates = this->dataPtr->distortionMap[distortedIdx] *
-                                     this->dataPtr->distortionTexWidth;
+          currDistortedCoordinates =
+            this->dataPtr->distortionMap[distortedIdx] *
+            this->dataPtr->distortionTexWidth;
 
           // grab new coordinates to map to
           newDistortedCoordinates[0] = mapCol;
@@ -504,7 +510,8 @@ ignition::math::Vector2d Distortion::Distort(
   dist.Y() += _p1 * (rSq + 2 * (normalized.Y()*normalized.Y())) +
       2 * _p2 * normalized.X() * normalized.Y();
 
-  return ((_center*_width) + ignition::math::Vector2d(dist.X(), dist.Y())*_f)/_width;
+  return ((_center*_width) +
+    ignition::math::Vector2d(dist.X(), dist.Y())*_f)/_width;
 }
 
 //////////////////////////////////////////////////
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -160,6 +160,10 @@ void Distortion::SetCamera(CameraPtr _camera)
   // seems to work best with a square distortion map texture
   unsigned int texSide = _camera->ImageHeight() > _camera->ImageWidth() ?
       _camera->ImageHeight() : _camera->ImageWidth();
+  // calculate focal length from largest fov
+  double fov = _camera->ImageHeight() > _camera->ImageWidth() ?
+      _camera->VFOV().Radian() : _camera->HFOV().Radian();
+  double focalLength = texSide/(2*tan(fov/2));
   this->dataPtr->distortionTexWidth = texSide - 1;
   this->dataPtr->distortionTexHeight = texSide - 1;
   unsigned int imageSize =
@@ -183,9 +187,9 @@ void Distortion::SetCamera(CameraPtr _camera)
       double u = j*incrU;
       ignition::math::Vector2d uv(u, v);
       ignition::math::Vector2d out =
-        this->Distort(uv, this->dataPtr->lensCenter,
+      this->Distort(uv, this->dataPtr->lensCenter,
             this->dataPtr->k1, this->dataPtr->k2, this->dataPtr->k3,
-            this->dataPtr->p1, this->dataPtr->p2);
+            this->dataPtr->p1, this->dataPtr->p2, this->dataPtr->distortionTexWidth, focalLength);
 
       // compute the index in the distortion map
       unsigned int idxU = out.X() * this->dataPtr->distortionTexWidth;
@@ -417,6 +421,35 @@ ignition::math::Vector2d Distortion::Distort(
       2 * _p2 * normalized.X() * normalized.Y();
 
   return _center + ignition::math::Vector2d(dist.X(), dist.Y());
+}
+
+//////////////////////////////////////////////////
+ignition::math::Vector2d Distortion::Distort(
+    const ignition::math::Vector2d &_in,
+    const ignition::math::Vector2d &_center, double _k1, double _k2, double _k3,
+    double _p1, double _p2, unsigned int _width, double _f)
+{
+  // apply Brown's distortion model, see
+  // http://en.wikipedia.org/wiki/Distortion_%28optics%29#Software_correction
+
+  ignition::math::Vector2d normalized2d = (_in - _center)*(_width/_f);
+  ignition::math::Vector3d normalized(normalized2d.X(), normalized2d.Y(), 0);
+  double rSq = normalized.X() * normalized.X() +
+               normalized.Y() * normalized.Y();
+
+  // radial
+  ignition::math::Vector3d dist = normalized * (1.0 +
+      _k1 * rSq +
+      _k2 * rSq * rSq +
+      _k3 * rSq * rSq * rSq);
+
+  // tangential
+  dist.X() += _p2 * (rSq + 2 * (normalized.X()*normalized.X())) +
+      2 * _p1 * normalized.X() * normalized.Y();
+  dist.Y() += _p1 * (rSq + 2 * (normalized.Y()*normalized.Y())) +
+      2 * _p2 * normalized.X() * normalized.Y();
+
+  return ((_center*_width) + ignition::math::Vector2d(dist.X(), dist.Y())*_f)/_width;
 }
 
 //////////////////////////////////////////////////
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -199,7 +199,31 @@ void Distortion::SetCamera(CameraPtr _camera)
           idxV < this->dataPtr->distortionTexHeight)
       {
         unsigned int mapIdx = idxV * this->dataPtr->distortionTexWidth + idxU;
-        this->dataPtr->distortionMap[mapIdx] = uv;
+
+        // if the destination pixel has been mapped previously,
+        // need to avoid the distorted image "overlapping" itself
+        ignition::math::Vector2d n1; // a bit of a hack since vector2d overloads operator= but not operator!=
+        n1 = -1;
+        if (this->dataPtr->distortionMap[mapIdx] != n1)
+        {
+         // grab original coordinates that map to this destination
+         ignition::math::Vector2d ij_ = this->dataPtr->distortionMap[mapIdx] * this->dataPtr->distortionTexWidth;
+         // grab new coordinates that want to map to this destination
+         ignition::math::Vector2d ij(i, j);
+         // grab center of the distortion
+         ignition::math::Vector2d uv_(this->dataPtr->lensCenter.X() * this->dataPtr->distortionTexWidth, 
+                                      this->dataPtr->lensCenter.Y() * this->dataPtr->distortionTexWidth);
+
+         // if new mapping is less extreme than former, use it instead
+         if (ij.Distance(uv_) < ij_.Distance(uv_))
+         {
+            this->dataPtr->distortionMap[mapIdx] = uv;
+         }
+        }
+        else
+        {
+          this->dataPtr->distortionMap[mapIdx] = uv;
+        }
       }
       // else: pixel maps outside the image bounds.
       // This is expected and normal to ensure
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -60,6 +60,12 @@ namespace gazebo
       /// black pixels at the corners of the image.
       public: bool distortionCrop = true;
 
+      /// \brief Modifies how Brown's distortion equations are applied to
+      /// better reflect real distortion. Image is projected from image plane
+      /// to camera plane to apply distortion equations, then projected back
+      /// to image plane. Note that this sets distortionCrop to false.
+      public: bool useRealDistortion = false;
+
       /// \brief Lens distortion compositor
       public: Ogre::CompositorInstance *lensDistortionInstance;
 
@@ -115,12 +121,20 @@ void Distortion::Load(sdf::ElementPtr _sdf)
   this->dataPtr->p2 = _sdf->Get<double>("p2");
   this->dataPtr->lensCenter = _sdf->Get<ignition::math::Vector2d>("center");
 
-  this->dataPtr->distortionCrop = this->dataPtr->k1 < 0;
-
   const std::string compositorName = "ignition:compositor";
   if (_sdf->HasElement(compositorName))
   {
     this->dataPtr->compositorName = _sdf->Get<std::string>(compositorName);
+  }
+  const std::string useRealDistortion = "ignition:use_real_distortion";
+  if (_sdf->HasElement(useRealDistortion))
+  {
+    this->dataPtr->useRealDistortion = _sdf->Get<bool>(useRealDistortion);
+  }
+  if (this->dataPtr->useRealDistortion) {
+    this->dataPtr->distortionCrop = false;
+  } else {
+    this->dataPtr->distortionCrop = this->dataPtr->k1 < 0;
   }
 }
 
@@ -204,13 +218,21 @@ void Distortion::SetCamera(CameraPtr _camera)
       normalizedLocation[0] = normalizedColLocation;
       normalizedLocation[1] = normalizedRowLocation;
 
-      distortedLocation = this->Distort(
-          normalizedLocation,
-          this->dataPtr->lensCenter,
-          this->dataPtr->k1, this->dataPtr->k2, this->dataPtr->k3,
-          this->dataPtr->p1, this->dataPtr->p2,
-          this->dataPtr->distortionTexWidth,
-          focalLength);
+      if (this->dataPtr->useRealDistortion) {
+        distortedLocation = this->Distort(
+            normalizedLocation,
+            this->dataPtr->lensCenter,
+            this->dataPtr->k1, this->dataPtr->k2, this->dataPtr->k3,
+            this->dataPtr->p1, this->dataPtr->p2,
+            this->dataPtr->distortionTexWidth,
+            focalLength);
+      } else {
+        distortedLocation = this->Distort(
+            normalizedLocation,
+            this->dataPtr->lensCenter,
+            this->dataPtr->k1, this->dataPtr->k2, this->dataPtr->k3,
+            this->dataPtr->p1, this->dataPtr->p2);
+      }
 
       // compute the index in the distortion map
       distortedCol = distortedLocation.X() * this->dataPtr->distortionTexWidth;
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -135,7 +135,9 @@ void Distortion::Load(sdf::ElementPtr _sdf)
   {
     this->dataPtr->legacyMode = _sdf->Get<bool>(legacyMode);
   }
-  if (!this->dataPtr->legacyMode) {
+  if (!this->dataPtr->legacyMode && this->dataPtr->distortionCrop)
+  {
+    gzwarn << "Enable legacy mode to use distortion cropping" << std::endl;
     this->dataPtr->distortionCrop = false;
   }
 }
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -168,8 +168,8 @@ void Distortion::SetCamera(CameraPtr _camera)
   this->dataPtr->distortionTexHeight = texSide - 1;
   unsigned int imageSize =
       this->dataPtr->distortionTexWidth * this->dataPtr->distortionTexHeight;
-  double incrU = 1.0 / this->dataPtr->distortionTexWidth;
-  double incrV = 1.0 / this->dataPtr->distortionTexHeight;
+  double colStepSize = 1.0 / this->dataPtr->distortionTexWidth;
+  double rowStepSize = 1.0 / this->dataPtr->distortionTexHeight;
 
   // initialize distortion map
   this->dataPtr->distortionMap.resize(imageSize);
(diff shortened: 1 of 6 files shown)


kjeppesen1:
@audrow

gazebo/rendering/Distortion.cc
```diff
@@ -186,7 +190,31 @@ void Distortion::SetCamera(CameraPtr _camera)
           idxV < this->dataPtr->distortionTexHeight)
       {
         unsigned int mapIdx = idxV * this->dataPtr->distortionTexWidth + idxU;
-        this->dataPtr->distortionMap[mapIdx] = uv;
+
+        // if the destination pixel has been mapped previously,
+        // need to avoid the distorted image "overlapping" itself
+        ignition::math::Vector2d n1; // a bit of a hack since vector2d overloads operator= but not operator!=
+        n1 = -1;
+        if (this->dataPtr->distortionMap[mapIdx] != n1)
```
kjeppesen1: There is probably a better way to do this; need to check if the value in this Vector2d object is equivalent to what it was set to initially (-1)

audrow:
One comment on the test we've added: We've added a more extreme barrel and pin cushion distortion to an existing test for distortion. This test is not very rigorous and may pass even before this PR or if the results are incorrect, for example because of parameter leaking https://github.com/osrf/gazebo/issues/2527.

I think after https://github.com/osrf/gazebo/issues/2527 is fixed, it would be good to add a stronger test to check that behavior is as expected. A possible work around for https://github.com/osrf/gazebo/issues/2527 is to make environments with one camera per world and test their expected behavior individually.

scpeters:
> This test is not very rigorous and may pass even before this PR or if the results are incorrect, for example because of parameter leaking #2527.

I have run this test without the fixes in this branch and confirmed that it does not fail, reducing its usefulness.

> for example because of parameter leaking #2527.

I believe that parameter leaking does not occur in this test because the individual cameras are each added dynamically using the `SpawnCamera` helper function. I think we don't need to wait for a fix to #2527 to improve this test and would prefer that we figure out how to do so in this pull request

scpeters:
> Before/after extreme distortion edge case fix:
> before fix after fix

how did you generate these images?

audrow:
> how did you generate these images?

We took them in Gazebo by placing a checkerboard directly in front of a camera with significant distortion (`k1 = -0.5`). Also, we're not doing border cropping here, so the gray around the edges is where there's no information.

scpeters:
this is a behavior change, so I think we should use a namespaced SDFormat parameter to opt-in to it, to keep existing camera parameters functioning the same way. Something like `<ignition:use_focal_length>`? Please suggest an appropriate parameter name. If you want to disable cropping with this flag, you can do that as well

~~~
<distortion>
  <ignition:use_focal_length>true</ignition:use_focal_length>
  <k1>...</k1>
  ...
</distortion>
~~~

kjeppesen1:
> this is a behavior change, so I think we should use a namespaced SDFormat parameter to opt-in to it, to keep existing camera parameters functioning the same way. Something like `<ignition:use_focal_length>`? Please suggest an appropriate parameter name.

I agree with this approach. Something that indicates that it will use the provided distortion coefficients directly would be ideal. Maybe `<ignition:use_real_distortion>`, `<ignition:use_exact_coefficients>`, or `<ignition:directly_use_coefficients>`, although those aren't great either.

>If you want to disable cropping with this flag, you can do that as well

I am in favor of disabling cropping with this implementation since cropping invalidates the distortion parameters (and therefore defeats the purpose of using the tag).

audrow:
@scpeters and @kjeppesen1, I've added a `<ignition:use_real_distortion>` tag and extended existing tests to make sure that this works. I also rebased this branch.

audrow:
> I believe that parameter leaking does not occur in this test because the individual cameras are each added dynamically using the `SpawnCamera` helper function. I think we don't need to wait for a fix to #2527 to improve this test and would prefer that we figure out how to do so in this pull request

I'll think about a better test to add.

audrow:
I've applied the feedback from yesterday and added  a test that I believe checks the feature.

Changes:
* Use `<ignition:legacy_mode>` flag
* Check for using legacy mode in the `setCrop` method
* Add a test that checks for expected cropping and distortion behavior with legacy and non legacy modes

I spawned cameras in the test environment to get around #2527.

gazebo/rendering/Distortion.cc
```diff
@@ -160,12 +173,16 @@ void Distortion::SetCamera(CameraPtr _camera)
   // seems to work best with a square distortion map texture
   unsigned int texSide = _camera->ImageHeight() > _camera->ImageWidth() ?
       _camera->ImageHeight() : _camera->ImageWidth();
+  // calculate focal length from largest fov
+  double fov = _camera->ImageHeight() > _camera->ImageWidth() ?
+      _camera->VFOV().Radian() : _camera->HFOV().Radian();
+  double focalLength = texSide/(2*tan(fov/2));
```
scpeters: nit: make these two variables (`fov`, `focalLength`) `const`

existing variables are fine; I'm just referring to these new variables

gazebo/rendering/Distortion.cc
```diff
@@ -174,30 +191,82 @@ void Distortion::SetCamera(CameraPtr _camera)
...
+      // Make sure the distorted pixel is within the texture dimensions
+      if (distortedCol < this->dataPtr->distortionTexWidth &&
+          distortedRow < this->dataPtr->distortionTexHeight)
       {
-        unsigned int mapIdx = idxV * this->dataPtr->distortionTexWidth + idxU;
-        this->dataPtr->distortionMap[mapIdx] = uv;
+        distortedIdx = distortedRow * this->dataPtr->distortionTexWidth + distortedCol;
+
+        // check if the index has already been set
+        if (this->dataPtr->distortionMap[distortedIdx] != unsetPixelVector)
```
scpeters, line 265: this logic block looks new. should we add `&& !this->dataPtr->legacyMode` to the if statement? or we could reverse it:

~~~
if (this->dataPtr->legacyMode)
{
  this->dataPtr->distortionMap[distortedIdx] = normalizedLocation;
}
// check if the index has already been set
else if (this->dataPtr->distortionMap[distortedIdx] != unsetPixelVector)
{
...
}
~~~

if not, then we should clarify that this is a bug fix that applies in all cases

gazebo/rendering/Distortion.cc
```diff
@@ -419,11 +488,40 @@ ignition::math::Vector2d Distortion::Distort(
   return _center + ignition::math::Vector2d(dist.X(), dist.Y());
```
scpeters: there's a bit of duplicate code here. Does it work to replace the body of the legacy `Distort` function with the following:

~~~
return Distort(_in, _center, _k1, _k2, _k3, _p1, _p2, 1u, 1.0);
~~~

I think that would simplify the diff and show what is different about the new function

test/integration/camera_sensor.cc
```diff
@@ -716,138 +722,202 @@ TEST_F(CameraSensor, CheckDistortion)
...
-  sensors::SensorPtr sensorUndistorted =
-    sensors::get_sensor(cameraNameUndistorted);
-  sensors::CameraSensorPtr camSensorUndistorted =
-    std::dynamic_pointer_cast<sensors::CameraSensor>(sensorUndistorted);
-  sensors::SensorPtr sensorFlat =
-    sensors::get_sensor(cameraNameFlat);
-  sensors::CameraSensorPtr camSensorFlat =
-    std::dynamic_pointer_cast<sensors::CameraSensor>(sensorFlat);
-  sensors::SensorPtr sensorBarrel =
+  for (auto isUseLegacyDistortionMode: {true, false}) {
```
scpeters: I'm skeptical of this `for` loop. I think it will work the first time through, but on the second pass, it will try to spawn models with the same name and pose but a different `legacyDistortionMode` value, which I don't think will work. What is your goal here?

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2110,231 @@ TEST_F(CameraSensor, SetCompositorNames)
     delete [] img;
   }
 #endif
-}
\ No newline at end of file
+}
+
+/////////////////////////////////////////////////
+TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
```
scpeters, line 2046: I haven't had a chance to look at the new test yet, but I'm excited to see it!

gazebo/rendering/Distortion.cc
audrow: Good call!

test/integration/camera_sensor.cc
audrow: The goal was to extend this test to cover the legacy and new modes. This was before I added the newest test. I didn't realize that they wouldn't spawn correctly. I can undo these changes, since I think the new distortion test is much more relevant.

gazebo/rendering/Distortion.cc
audrow: I think this can be considered a bug fix, since there will be the wrapping behavior shown in https://github.com/osrf/gazebo/pull/3009#issue-650530018 on very large barrel distortions with the legacy method.

I'm not sure what would be the best place to mention this. I can alter the PR to mention the bug fix and add a comment in this part of the code. Anywhere else?

test/integration/camera_sensor.cc
audrow: I removed these changes from the PR - the new test is still included.

gazebo/rendering/Distortion.cc
audrow: I added a comment in the code (below) and modified the PR title and description. https://github.com/osrf/gazebo/blob/406b65e634b3ee6d45ffbd6f77d136caa048ca3a/gazebo/rendering/Distortion.cc#L240-L243

gazebo/test/ServerFixture.hh
```diff
@@ -254,6 +254,7 @@ namespace gazebo
     /// \param[in] _distortionP2 Distortion coefficient p2.
     /// \param[in] _cx Normalized optical center x, used for distortion.
     /// \param[in] _cy Normalized optical center y, used for distortion.
+    /// \param[in] _useRealDistortion Use real distortion
```
scpeters: need to fix this comment to use `_legacyMode` and update the description

gazebo/rendering/Distortion.cc
```diff
@@ -122,6 +130,11 @@ void Distortion::Load(sdf::ElementPtr _sdf)
   {
     this->dataPtr->compositorName = _sdf->Get<std::string>(compositorName);
   }
+  const std::string legacyMode = "ignition:legacy_mode";
+  if (_sdf->HasElement(legacyMode))
+  {
+    this->dataPtr->legacyMode = _sdf->Get<bool>(legacyMode);
```
scpeters, line 136: just to be safe, I would set `this->dataPtr->distortionCrop = false;` here as well

scpeters (commented):
thanks for applying those changes; I just noticed two small things, and I'll look at the new test next week

gazebo/rendering/Distortion.cc
```diff
@@ -122,6 +130,11 @@ void Distortion::Load(sdf::ElementPtr _sdf)
   {
     this->dataPtr->compositorName = _sdf->Get<std::string>(compositorName);
   }
+  const std::string legacyMode = "ignition:legacy_mode";
+  if (_sdf->HasElement(legacyMode))
```
kjeppesen1, line 134: A debug-level message somewhere, possibly here, might be helpful to inform the user what mode is active. That would inform people who kept the default that this other implementation exists.

gazebo/rendering/Distortion.cc
```diff
@@ -417,7 +417,8 @@ void Distortion::SetCamera(CameraPtr _camera)
   this->dataPtr->lensDistortionInstance->getTechnique()->getOutputTargetPass()->
       getPass(0)->setMaterial(this->dataPtr->distortionMaterial);
 
-  this->CalculateAndApplyDistortionScale();
+  if (this->dataPtr->legacyMode)
```
kjeppesen1: I just added this fix. Distortion was still being applied when legacyMode was false since this CalculateAndApplyDistortionScale() was always being called. Cropping now seems to behave as intended.

test/integration/camera_sensor.cc
```diff
@@ -40,10 +40,14 @@ unsigned char* img = NULL;
 unsigned char* img2 = NULL;
 unsigned char* img3 = NULL;
 unsigned char* img4 = NULL;
+unsigned char* img5 = NULL;
```
audrow: Hey @kjeppesen1, it looks like c3f5cf7 didn't make any changes and 464adb1 reapplied some old changes (e.g., I removed my additions to this test, among other things). We could cherry-pick 4abe770 on top of 406b65e? I can probably do this tomorrow, as well as address @scpeters's feedback.

gazebo/rendering/Distortion.cc
scpeters: @kjeppesen1 ok maybe my comment here is necessary, not "just to be safe"

I think I prefer this approach over 4abe770

audrow:
@mvespign, thanks for removing the duplicate method. This was done by https://github.com/osrf/gazebo/pull/3009/commits/4fe9c4b75b958c6e0e3481bd027bfa7fda1dd8ae (before https://github.com/osrf/gazebo/pull/3009/commits/464adb195cf20b495c76b65a0199e097138fda0f). I'm going to force push to remove the recent commits.

gazebo/rendering/Distortion.cc
audrow: I added a warning for this :+1:

gazebo/rendering/Distortion.hh
```diff
@@ -103,6 +104,24 @@ namespace gazebo
...
+      /// \param[in] _p1 Tangential distortion coefficient p1.
+      /// \param[in] _p2 Tangential distortion coefficient p2.
+      /// \param[in] _width Width of the image texture in pixels.
+      /// \param[in] _f Focal length in pixels.
+      /// \return Distorted coordinate.
+      public: static ignition::math::Vector2d Distort(
+                  const ignition::math::Vector2d &_in,
+                  const ignition::math::Vector2d &_center, 
+                  double _k1, double _k2, double _k3,
+                  double _p1, double _p2,
```
scpeters: there's some trailing whitespace in this file; can you clean it up?

worlds/test/issue_3003_distortion_implementation_correction.world
```diff
@@ -0,0 +1,65 @@
...
+      <ambient>0.4 0.4 0.4 1</ambient>
+      <background>1 1 1 1</background>
+    </scene>
+    <model name='minecart'>
+      <static>1</static>
+      <link name='base_link'>
+        <collision name='collision'>
+          <geometry>
+            <mesh>
+              <uri>/root/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/mine cart/4/meshes/minecart.dae</uri>
```
scpeters: these paths with `/root` are not valid on my machine. I recommend the following instead:

~~~
https://fuel.ignitionrobotics.org/1.0/openrobotics/models/mine cart/4/files/meshes/minecart.dae
~~~

worlds/test/issue_3003_distortion_implementation_correction.world
```diff
@@ -0,0 +1,65 @@
...
+                <ode/>
+              </torsional>
+              <ode/>
+            </friction>
+          </surface>
+        </collision>
+        <visual name='polySurface1278_visual'>
+          <geometry>
+            <mesh>
+              <uri>/root/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/mine cart/4/meshes/minecart.dae</uri>
```
scpeters: `https://fuel.ignitionrobotics.org/1.0/openrobotics/models/mine cart/4/files/meshes/minecart.dae`

worlds/test/issue_3003_distortion_implementation_correction.world
```diff
@@ -0,0 +1,65 @@
...
+            <pbr>
+              <metal>
+                <albedo_map>materials/textures/MineCart_Albedo.png</albedo_map>
+                <metalness_map>materials/textures/MineCart_Metalness.png</metalness_map>
+                <roughness_map>materials/textures/MineCart_Roughness.png</roughness_map>
+              </metal>
+            </pbr>
+            <script>
+              <uri>/root/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/mine cart/4/materials/scripts/</uri>
+              <uri>/root/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/mine cart/4/materials/textures/</uri>
```
scpeters: `https://fuel.ignitionrobotics.org/1.0/openrobotics/models/mine cart/4/files/materials/textures/`

worlds/test/issue_3003_distortion_implementation_correction.world
```diff
@@ -0,0 +1,65 @@
...
+            <specular>1 1 1 1</specular>
+            <pbr>
+              <metal>
+                <albedo_map>materials/textures/MineCart_Albedo.png</albedo_map>
+                <metalness_map>materials/textures/MineCart_Metalness.png</metalness_map>
+                <roughness_map>materials/textures/MineCart_Roughness.png</roughness_map>
+              </metal>
+            </pbr>
+            <script>
+              <uri>/root/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/mine cart/4/materials/scripts/</uri>
```
scpeters: `https://fuel.ignitionrobotics.org/1.0/openrobotics/models/mine cart/4/files/materials/scripts/`

worlds/test/issue_3003_distortion_implementation_correction.world
audrow: Thanks for pointing this out.

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,230 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+    unsigned int pixelSum = r4 + g4 + b4;
+    if (pixelSum != 765u && pixelSum != 192u) {
+      idxOfFirstDumpsterPixelImg4 = x;
+      break;
+    }
+  }
+  // Check that, in the pin cushion case, the dumpster is seen closer to the
+  // left in the new version. This makes sense as the new distortion mode
+  // should have more significant distortion than the legacy mode.
+  ASSERT_GT(idxOfFirstDumpsterPixelImg1, idxOfFirstDumpsterPixelImg2+15);
```
scpeters: nit: use `EXPECT_GT` instead of `ASSERT_GT`

I mainly use `ASSERT_*` calls to ensure that pointers are not null before dereferencing; this is not such a case

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,230 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+  // Check that, in the barrel case, the dumpster is seen closer to the left
+  // in the legacy distortion mode than in the new mode. This makes sense
+  // because the legacy mode crops the image removing the edges where there
+  // is no image information.
+  ASSERT_LT(idxOfFirstDumpsterPixelImg3, idxOfFirstDumpsterPixelImg4-25);
```
scpeters: nit: `EXPECT_LT` instead of `ASSERT_LT`

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,230 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+  // Compare colors. Barrel distorted image should have more darker pixels than
+  // the original as the ground plane has been warped to occupy more of the
+  // image. The same should be true for pincushion distortion, because the
+  // ground plane is still distorted to be larger - just different parts
+  // of the image are distorted.
+  unsigned int colorSum1 = 0;
+  unsigned int colorSum2 = 0;
+  unsigned int colorSum3 = 0;
+  unsigned int colorSum4 = 0;
+  unsigned int dumpsterRow = height / 2;
```
scpeters: nit: make this `const`

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,230 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+  // Make sure the render engine is available.
+  if (rendering::RenderEngine::Instance()->GetRenderPathType() ==
+      rendering::RenderEngine::NONE)
+  {
+    gzerr << "No rendering engine, unable to run camera test\n";
+    return;
+  }
+
+  // Spawn cameras.
+  std::string modelNameBarrelLegacy = "camera_model_barrel_legacy";
```
scpeters: nit: make these parameters `const` from this down to `horizontalFov`

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,230 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+    imageCount3 < 10 || imageCount4 < 10)
+  {
+    common::Time::MSleep(10);
+  }
+
+  // Compare colors. Barrel distorted image should have more darker pixels than
+  // the original as the ground plane has been warped to occupy more of the
+  // image. The same should be true for pincushion distortion, because the
+  // ground plane is still distorted to be larger - just different parts
+  // of the image are distorted.
```
scpeters: I agree with the content of this comment block, but we aren't actually testing this, since we don't have an undistorted image to compare against

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,230 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+    unsigned int g3 = img3[(dumpsterRow*width*3) + x + 1];
+    unsigned int b3 = img3[(dumpsterRow*width*3) + x + 2];
+    colorSum3 += r3 + g3 + b3;
+    unsigned int r4 = img4[(dumpsterRow*width*3) + x];
+    unsigned int g4 = img4[(dumpsterRow*width*3) + x + 1];
+    unsigned int b4 = img4[(dumpsterRow*width*3) + x + 2];
+    colorSum4 += r4 + g4 + b4;
+  }
+
+  // Check that the legacy mode distorts the pincuchin images less
```
scpeters: pincuchin -> pincushion

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,230 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+  EXPECT_GT(colorSum1, colorSum2+800);
+  // Check that there is a good difference in the barrel images
+  // this difference is largely caused by the edges of the new
+  // distortion model which appear gray when they have no value
+  // and gray (192) has a much lower value than white (765)
+  EXPECT_GT(colorSum3, colorSum4+20000);
+
+  // Check that no cropping occurs - specifically the corners
```
scpeters: > "Check that no cropping occurs"

I think this isn't quite right. The legacy barrel image is actually cropped, which is why its corner pixel is white. The pincushion images isn't cropped. Maybe "Check that the corners contain no unrendered pixels"?

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,230 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+  // since the background of the environment is white
+  unsigned int cornerColorSumImg1 = img[0] + img[1] + img[2];
+  unsigned int cornerColorSumImg2 = img2[0] + img2[1] + img2[2];
+  unsigned int cornerColorSumImg3 = img3[0] + img3[1] + img3[2];
+  EXPECT_EQ(cornerColorSumImg1, 765u);
+  EXPECT_EQ(cornerColorSumImg2, 765u);
+  EXPECT_EQ(cornerColorSumImg3, 765u);
+  // Check that cropping occurs - specifically the corners
+  // in both image should be white and have a value of 192,
+  // which is the gray for pixels that haven't been assigned
```
scpeters: > Check that cropping occurs - specifically the corners in both image should be white

Likewise, this isn't quite right. This image is *not* cropped, and the corner pixels are not white; they are gray, which is used for unrendered pixels. Perhaps "Check that this image is not cropped and that the corner pixel has the gray value assigned to unrendered pixels"

scpeters (commented):
the new test looks good. I think you've matched the style of other tests, though there is a fair bit of repeated code. Some lambda functions could probably reduce some code duplication, but it's fine as is

test/integration/camera_sensor.cc
audrow: Good catch!

test/integration/camera_sensor.cc
audrow: D'oh!

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,196 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+    {
+      const unsigned int r = img[(middleRow*width*3) + x];
+      const unsigned int g = img[(middleRow*width*3) + x + 1];
+      const unsigned int b = img[(middleRow*width*3) + x + 2];
+      const unsigned int pixelSum = r + g + b;
+      if (pixelSum != 765u && pixelSum != 192u) {
+        return x;
+      }
+    }
+    return width-1;
```
scpeters: nit: `return (width-1) * 3;`

scpeters (commented):
ok, I think this is almost there. I just have one minor nit, and I noticed some lines are longer than 80 characters. Our CI is not currently reporting style errors due to https://github.com/ignition-tooling/release-tools/issues/446, but I think avoiding trailing whitespace and long lines should be enough

test/integration/camera_sensor.cc
audrow: Good catch!

audrow:
I think that it should be good now.

scpeters:
I've tested with libgazebo_ros_camera, and I think the logic is working properly now

test/integration/camera_sensor.cc
```diff
@@ -2040,4 +2040,198 @@ TEST_F(CameraSensor, SetCompositorNames)
...
+  unsigned int cornerColorSumImg3 = img3[0] + img3[1] + img3[2];
+  EXPECT_EQ(cornerColorSumImg1, 765u);
+  EXPECT_EQ(cornerColorSumImg2, 765u);
+  EXPECT_EQ(cornerColorSumImg3, 765u);
+  // Check that this image is not cropped and that the corner pixel
+  // has the gray value assigned to unrendered pixels
+  unsigned int cornerColorSumImg4 = img4[0] + img4[1] + img4[2];
+  EXPECT_EQ(cornerColorSumImg4, 192u);
+
+  auto getFirstColIdxOfMineCart = [middleRow, width](const unsigned char* img)
```
scpeters: I just noticed that clang warns that you don't need to capture `middlerow` and `width`

* https://build.osrfoundation.org/job/gazebo-ci-pr_any-homebrew-amd64/2443/clang/folder.-566429772/fileName.-806121611/

I'm guessing it is because they are `const` integers?

scpeters (approved):
I just noticed some new clang warnings, but aside from that I approve

gazebo/rendering/Distortion.cc
```diff
@@ -174,30 +196,93 @@ void Distortion::SetCamera(CameraPtr _camera)
     this->dataPtr->distortionMap[i] = -1;
   }
 
+  ignition::math::Vector2d distortionCenterCoordinates(
+      this->dataPtr->lensCenter.X() * this->dataPtr->distortionTexWidth,
+      this->dataPtr->lensCenter.Y() * this->dataPtr->distortionTexWidth);
+
+  // declare variables before the loop
+  static auto unsetPixelVector =  ignition::math::Vector2d(-1, -1);
```
iche033: does this need to be `static`? maybe just `const`?

gazebo/rendering/Distortion.cc
```diff
@@ -174,30 +196,93 @@ void Distortion::SetCamera(CameraPtr _camera)
...
-      double u = j*incrU;
-      ignition::math::Vector2d uv(u, v);
-      ignition::math::Vector2d out =
-        this->Distort(uv, this->dataPtr->lensCenter,
+      normalizedColLocation = mapCol*colStepSize;
+
+      normalizedLocation[0] = normalizedColLocation;
+      normalizedLocation[1] = normalizedRowLocation;
+
+      if (this->dataPtr->legacyMode) {
```
iche033: style nitpick: `{` on new line

gazebo/rendering/Distortion.cc
```diff
@@ -174,30 +196,93 @@ void Distortion::SetCamera(CameraPtr _camera)
...
+        distortedLocation = this->Distort(
+            normalizedLocation,
+            this->dataPtr->lensCenter,
             this->dataPtr->k1, this->dataPtr->k2, this->dataPtr->k3,
             this->dataPtr->p1, this->dataPtr->p2);
+      } else {
```
iche033: style nitpick: `{` on new line

gazebo/test/ServerFixture.hh
```diff
@@ -267,7 +268,9 @@ namespace gazebo
                    bool _distortion = false, double _distortionK1 = 0.0,
                    double _distortionK2 = 0.0, double _distortionK3 = 0.0,
                    double _distortionP1 = 0.0, double _distortionP2 = 0.0,
-                   double _cx = 0.5, double _cy = 0.5);
+                   double _cx = 0.5, double _cy = 0.5,
+                   bool _legacyMode = true,
+                   double _horizontalFov = 0.78539816339744828);
```
iche033, line 274: I think this breaks ABI? 

ABI checker didn't show any issue here but I'm not sure if that's because it's not testing the `ServerFixture` library. 

Also add `\param[in]` for `_horizontalFov`

scpeters: > I think this breaks ABI?
> 
> ABI checker didn't show any issue here but I'm not sure if that's because it's not testing the `ServerFixture` library.

I was wondering the same thing, but we only distribute `ServerFixture` in a static library, and I believe ABI is only a concern for dynamically loaded libraries

iche033 (approved):

scpeters:
Windows CI failed for an unrelated reason, and the failures in the other jobs look unrelated to me. Merging now...

PR #3013, opened 2021-05-26
Build gazebo_test_fixture target when running make install

j-rivero:
When running `make install` without other commands `gazebo_test_fixture` is not build since no target needs it. That is triggering a failure since the install target will look for the library to be installed ([see Debian log](https://buildd.debian.org/status/fetch.php?pkg=gazebo&arch=all&ver=11.5.1%2Bdfsg-1&stamp=1621980186&file=log)).

The PR adds the target to the ALL target which is being used by install.

### DIFF
--- a/gazebo/test/CMakeLists.txt
+++ b/gazebo/test/CMakeLists.txt
@@ -8,3 +8,5 @@ gz_install_includes("test"
   ServerFixture.hh
   ${PROJECT_BINARY_DIR}/test_config.h
 )
+
+set_target_properties(gazebo_test_fixture PROPERTIES DEPENDS ALL)


scpeters:
I'm not able to reproduce the failure when running `make install` or `make preinstall` (which I noticed in the debian log) right after cmake. This seems like a small change, but I'd like to be sure that it fixes the problem we are seeing in debian

j-rivero:
> I'm not able to reproduce the failure when running `make install` or `make preinstall` (which I noticed in the debian log) right after cmake. This seems like a small change, but I'd like to be sure that it fixes the problem we are seeing in debian

You are right of both, the make install did not triggered the problem on Gazebo vanilla code and the patch does not fix the problems in Debian. Should be something else (probably in Debian patches) that I need to detect. Thanks Steve.

PR #3030, opened 2021-06-21
Use CURL::libcurl instead of variables

sloretz:
Maybe this will fix the build failures in this job :shrug: 
https://build.osrfoundation.org/job/gazebo-ci-pr_any-windows7-amd64/ws/ws/

The trouble is `gazebo_common` is linking against `libcurl.dll` instead of `libcurl_imp.lib`. I have no idea why that happens. The exported CMake targets on Windows seem reasonable to me, so maybe using them will solve the issue.

### DIFF
--- a/gazebo/common/CMakeLists.txt
+++ b/gazebo/common/CMakeLists.txt
@@ -17,10 +17,6 @@ if (HAVE_GDAL)
   include_directories(${GDAL_INCLUDE_DIR})
 endif()
 
-if (CURL_FOUND AND WIN32)
-  add_definitions(-DCURL_STATICLIB)
-endif()
-
 include_directories(${tinyxml_INCLUDE_DIRS})
 link_directories(${tinyxml_LIBRARY_DIRS})
 
--- a/gazebo/common/CMakeLists.txt
+++ b/gazebo/common/CMakeLists.txt
@@ -229,7 +229,6 @@ target_compile_definitions(gazebo_common
 target_link_libraries(gazebo_common
   PUBLIC
     ${Boost_LIBRARIES}
-    CURL::libcurl
     ${freeimage_LIBRARIES}
     ignition-common3::graphics
     ${IGNITION-FUEL_TOOLS_LIBRARIES}
@@ -240,6 +239,7 @@ target_link_libraries(gazebo_common
     ${SDFormat_LIBRARIES}
     ${tinyxml_LIBRARIES}
   PRIVATE
+    CURL::libcurl
     ${libtool_library}
     ${libavcodec_LIBRARIES}
     ${libavformat_LIBRARIES}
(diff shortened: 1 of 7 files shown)


scpeters:
looks like it works on windows but not Ubuntu bionic

* [Build Status](https://build.osrfoundation.org/job/gazebo-ci-pr_any-windows7-amd64/3259/) https://build.osrfoundation.org/job/gazebo-ci-pr_any-windows7-amd64/3259/
* [Build Status](https://build.osrfoundation.org/job/gazebo-ci-pr_any-ubuntu_auto-amd64-gpu-nvidia/688/) https://build.osrfoundation.org/job/gazebo-ci-pr_any-ubuntu_auto-amd64-gpu-nvidia/688/


~~~
CMake Error at cmake/GazeboUtils.cmake:99 (add_library):
  Target "libgazebo" links to target "CURL::libcurl" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
Call Stack (most recent call first):
  gazebo/CMakeLists.txt:108 (gz_add_library)
~~~

gazebo/common/CMakeLists.txt
```diff
@@ -237,7 +229,7 @@ target_compile_definitions(gazebo_common
 target_link_libraries(gazebo_common
   PUBLIC
     ${Boost_LIBRARIES}
-    ${CURL_LIBRARIES}
+    CURL::libcurl
```
traversaro: Not something introduced in this PR, but it seems that `CURL` is not used in public headers, so probably `CURL::libcurl` can be linked as a `PRIVATE` linked library, instead of a `PUBLIC` one.

sloretz: Made PRIVATE in 9057ef9

tools/CMakeLists.txt
```diff
@@ -53,6 +45,7 @@ if (WIN32)
 endif()
 
 target_link_libraries(gz
+ CURL::libcurl
```
scpeters: I don't see any usage of `curl` by `gz`; I would remove this line

I'm guessing the cmake code you've removed in this file is vestigial

scpeters (commented):
I think most of the plugins don't need to link to `curl`; I only see 3 that should need it:

~~~
grep -rnI include\ .curl/ plugins
plugins/rest_web/RestUiLoginDialog.cc:19:#include <curl/curl.h>
plugins/rest_web/RestApi.cc:20:#include <curl/curl.h>
plugins/StaticMapPlugin.cc:18:#include <curl/curl.h>
~~~

In fact, I think we could remove the `curl` include from RestUiLoginDialog.cpp, since it doesn't actually use it for anything. I'll try testing some changes and suggest a patch

tools/CMakeLists.txt
scpeters: dba62f3

scpeters:
I just pushed my changes directly; hope you don't mind

sloretz:
All CI jobs appear to have built :tada: . I don't see any new CMake warnings. As for test failures, there are lots, but that doesn't seem unusual looking at the past jobs. Anything left to do before this can be merged?

scpeters (approved):
looks good to me

PR #3031, opened 2021-06-23
Make links within nested models modifiable from GUI Client

lihui815:
Links within (properly) nested models can't be modified:

- `World::ModelById()` calls `Base::getById()`, which is not recursive (one layer deep only), so it can't find nested models to modify on a `~/model/modify` request. Create a new method `Base::GetByIdRecursive()` in that searches for descendent entities in the tree recursively, and use it in `World::ModelById()` .

Once ^^ is solved, it will expose that model and link names get reset on a modify erroneously:

- On receiving a `~/model/modify` request, the model name in the request message has `StripWorldName()` applied to it, and then the result is used to reset the model name. This is assuming that the model is a top-element of the world and cannot be nested. Changed this to use new method `StripParentScopeName()` to properly get the short (unscoped) requesting name.
- On sending a `~/model/modify` request, the client sends the short (unscoped) link name and the server expects the short (unscoped) link name to set the link name. However, the server then publishes the request message back as the `~/model/info` for the response to the client. The client is actually expecting full (scoped) names for everything, thus this change also includes sending the full (scoped) name for the link
- ^^ by itself will cause link names to be renamed erroneously on the server side, because the server is resetting short (unscoped) link names to the link name in the `~/model/modify` request. Therefore, this change also to applies `StripParentScopeName()` to the link name in a modify request on the server side.

Also added tests for:

- link properties for worlds with nested models


emersonknapp:
@j-rivero @chapulina friendly ping, can you assign a reviewer for this one? thanks!

iche033:
I verified that modifying nested model/link properties though the ModelListWidget on the left hand pane now works as expected.

I made a few changes in the [iche033/gazebo11-fn](https://github.com/osrf/gazebo/commit/142894877e482a32cfee5dd3f7b8b6daf6fd4fb9) branch:
* style fixes
* moved `multilink` models to tthe est directory as they are only used for tests
* fixed flaky `PluginProperties()` test in `ModelListWidget_TEST`
* added unit test for `Strip*ScopedName` functions

Can you take a look and merge if they look good to you?

lihui815:
@iche033 thanks, merged and resubmitted.

iche033 (approved):
looks good to me

PR #3032, opened 2021-06-24
Install Profiler header

eborghi10:
In order to use the Profiler in custom plugins, the header must be installed.

```c++
#include <gazebo/common/Profiler.hh>
```

This PR installs `Profiler.hh` when `ENABLE_PROFILER=1`.

### DIFF
--- a/gazebo/common/CMakeLists.txt
+++ b/gazebo/common/CMakeLists.txt
@@ -226,6 +226,10 @@ set (remotery_sources
   RemoteryProfilerImpl.cc
 )
 
+set(profiler_headers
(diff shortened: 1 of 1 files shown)


scpeters (approved):

j-rivero:
Great! Could you please include a Changelog entry?

scpeters:
> Great! Could you please include a Changelog entry?

let's handle the Changelog when we're ready for a release; I think it doesn't need it here

chapulina (approved):
Great to see the profiler is being handy :+1:

PR #3036, opened 2021-06-30
Enable hardware PCF for spotlight shadows

WilliamLewww:
The `FocusedShadowCameraSetup` causes linear artifacts that was fixed by changing the shadow camera setup to `DefaultShadowCameraSetup`. There is a decrease in quality for shadows generated with the `DefaultShadowCameraSetup` but it shouldn't be a problem when hardware PCF is enabled.

### DIFF
--- a/gazebo/rendering/Light.cc
+++ b/gazebo/rendering/Light.cc
@@ -588,7 +588,7 @@ void Light::SetCastShadows(const bool _cast)
     if (_cast && this->dataPtr->shadowCameraSetup.isNull())
     {
       this->dataPtr->shadowCameraSetup =
-          Ogre::ShadowCameraSetupPtr(new Ogre::FocusedShadowCameraSetup());
+          Ogre::ShadowCameraSetupPtr(new Ogre::DefaultShadowCameraSetup());
       this->dataPtr->light->setCustomShadowCameraSetup(
           this->dataPtr->shadowCameraSetup);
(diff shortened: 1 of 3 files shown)


iche033:
can you update the [shader](https://github.com/osrf/gazebo/blob/gazebo11/media/materials/programs/spotlight_shadow_demo_fp.glsl) used by the [spotlight_shadows_demo.world](https://github.com/osrf/gazebo/blob/gazebo11/worlds/spotlight_shadows_demo.world) to demo the new spotlight shadows with hardware PCF? Currently with the changes in this PR, the shadows no longer get rendered.

WilliamLewww:
I forgot that there was an Ogre change that might go along with the pull request. I'll put this pull request in draft and open it back up once I find a workaround or if the Ogre change has no effect on this pull request.

WilliamLewww:
For some reason the shadows are working fine for me with the changes.

a

@scpeters If possible, could you try the branch and run the `spotlight_shadows_demo.world`? Hopefully it isn't an issue regarding hardware.

iche033 (approved):
works for me now.

scpeters:
I'm going to merge this with `gazebo11` to rerun CI since the ABI checker failed

scpeters (approved):
macOS only supports version 120 of glsl, so the `spotlight_shadow_demo_fp.glsl` doesn't work on macOS, but that's not a blocker

PR #3039, opened 2021-07-07
Partially fixes #1718 - Apply Force/Torque for nested models

### LINKED ISSUE #1718: Nested model issues
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/1718)) by Ian Chen (Bitbucket: [Ian Chen](https://bitbucket.org/%7Beaa6fca5-6deb-43f6-907f-971c144735dd%7D/), GitHub: [iche033](https://github.com/iche033)).**

[pull request #1868](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/1868) started work on nested model. 

Just some notes about what currently works for nested models and what doesn't:

- Logging: seems to record and playback correctly

- Editing: Missing nested models when loading in the editor

- Joint control panel: seems to work ok

- Model list widget: nested models don't show under parent model as they should.

- View collisions, CoM and so on: ok

- Inserting:

    - Preview doesn't load nested models .

    - Nested model doesn't show on model list at all.

    - Insert two instances of the same model - nested models don't collide with each other

- Copy / pasting: pasted nested model doesn't collide with original nested model but collides with other models.

- Align: ok

- Snap: ok

- Apply force/torque: nested model shows on the list and applying force to that doesn't work.

- Context menu: Right-clicking a nested model selects the parent model.

- Deleting: ok

- Physics engines: 

    - ODE: yes

    - bullet: yes

    - Simbody:  No
    
    - DART: (need to confirm)

camm73:
The following fixes were applied due to issues encountered when selecting `Apply Force/Torque` for nested models and its links from `ModelRightMenu`:
- `ModelRightMenu::Run()` previously set the `entityName` for nested models to be the model name before the first `::` delimiter, or the top-most parent model. This incorrectly initializes ApplyWrenchDialog for the top-most parent model rather than the nested model selected by the user. This was solved by setting `entityName` to the fully scoped name for all models.
- `ModelRightMenu::onApplyWrench()` previously made the assumption that the Visual for all models would be the RootVisual, however this is not the case for nested models. This was fixed by instead distinguishing models from links by their visual type: `Visual::VT_MODEL`.
- Unscoped link names within `ApplyWrenchDialog` were previously generated using `std::string::find()` with the `::` delimiter; however, this only removes the top-most parent model in nested models, leaving a partially-scoped link name. This is resolved by the using `std::string::find()` to locate the start of the model name within the scoped link name and then extracting the sub-string following the model name as the unscoped link name.
- The ComboBox for Links in the `ApplyWrenchDialog` previously included nested models within the list. This was fixed by type checking each Visual when generating the ComboBox to ensure each element is of type: `Visual::VT_LINK`.

The following changes were made for testing:
- A new world `nested_multilink_shape.world` was created to properly test the changes made. The existing "nested_model.world" could not be used since each model only has a single link, not allowing selecting multiple links from the Link ComboBox in `ApplyWrenchDialog`. The "multilink_shape.world" could also not be updated to add another nested model since other tests for `ModelListWidget` rely on it having only 1 model.
- The `ApplyWrenchDialog_TEST::ApplyForceTorqueFromDialog()` test was updated to use the `nested_multilink_shape.world` and was configured with a for-loop to apply the same exact tests to the nested model as are applied to the original model.

Tests were **not** included for the changes to `ModelRightMenu` since this class was previously un-tested and the effects of the code changes reside in private functions that are connected to private `QAction`s. In addition, the `QMenu` in which these actions reside is a local variable without a parent `QWidget` and is therefore inaccessible from any unit testing files. However, the formatting of the model name and link name in `ModelRightMenu::onApplyWrench()` can be seen to be of the same format as used for direct testing in `ApplyWrenchDialog_TEST::ApplyForceTorqueFromDialog()`.

### DIFF
--- a/gazebo/gui/ApplyWrenchDialog.cc
+++ b/gazebo/gui/ApplyWrenchDialog.cc
@@ -488,7 +488,8 @@ bool ApplyWrenchDialog::SetModel(const std::string &_modelName)
     rendering::VisualPtr childVis = vis->GetChild(i);
 
     // Skip all children that aren't links
-    if(childVis->GetType() != rendering::Visual::VT_LINK){
+    if (childVis->GetType() != rendering::Visual::VT_LINK)
+    {
       continue;
     }
 
--- a/gazebo/gui/ApplyWrenchDialog.cc
+++ b/gazebo/gui/ApplyWrenchDialog.cc
@@ -486,6 +486,12 @@ bool ApplyWrenchDialog::SetModel(const std::string &_modelName)
   for (unsigned int i = 0; i < vis->GetChildCount(); ++i)
   {
     rendering::VisualPtr childVis = vis->GetChild(i);
+
+    // Skip all children that aren't links
+    if(childVis->GetType() != rendering::Visual::VT_LINK){
+      continue;
+    }
+
     std::string linkName = childVis->Name();
 
     // Issue #1553: This is failing to get real links sometimes:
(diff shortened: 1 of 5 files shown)


iche033:
changes look good. I was able to trigger the `Apply Force / Torque` dialog on the inner nested model's links by right clicking on the inner nested model through the entity tree in the left panel. 

Right clicking directly on one of the inner nested models' links does not work since that's selecting the outer inner model. I think mouse selection for nested model links could be improved but that'll be outside the scope of this PR.

iche033 (approved):

Issue #2874, opened 2020-11-06
Gazebo9 - Race condition between EventT::Signal and EventT::Connect leading to segfault during world update


ojasjoshi:
Stack trace:
```bash
#0  std::__atomic_base<bool>::load (__m=std::memory_order_seq_cst, this=0x0) at /usr/include/c++/5/bits/atomic_base.h:396
#1  std::atomic<bool>::operator bool (this=0x0) at /usr/include/c++/5/atomic:81
#2  gazebo::event::EventT<void (gazebo::common::UpdateInfo const&)>::Signal<gazebo::common::UpdateInfo>(gazebo::common::UpdateInfo const&) (
    this=0x7f60bc054940 <gazebo::event::Events::worldUpdateBegin>, _p=...) at /gazebo/gazebo/common/Event.hh:384
#3  0x00007f60ba9aee59 in gazebo::event::EventT<void (gazebo::common::UpdateInfo const&)>::operator()<gazebo::common::UpdateInfo>(gazebo::common::UpdateInfo const&) (_p=..., this=<optimized out>) at /gazebo/gazebo/common/Event.hh:221
#4  gazebo::physics::World::Update (this=this@entry=0x274a180) at /gazebo/gazebo/physics/World.cc:690
#5  0x00007f60ba9be4df in gazebo::physics::World::Step (this=this@entry=0x274a180) at /gazebo/gazebo/physics/World.cc:621
#6  0x00007f60ba9be955 in gazebo::physics::World::RunLoop (this=0x274a180) at /gazebo/gazebo/physics/World.cc:469
#7  0x00007f60b844c5d5 in ?? ()
   from /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.58.0
#8  0x00007f60bace36ba in start_thread (arg=0x7f5fb4bf4700) at pthread_create.c:333
#9  0x00007f60bb2ed41d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

A call to `EventT::Connect` from any subscriber creates a `new EventConnection(true, _subscriber)` between the event and the subscriber.  A call to `Event::Signal` during `worldUpdateBeginLoop` then executes callbacks to all the `EventConnection`'s added to this event. 

A race condition occurs between `EventT::Signal` and `EventT::Connect` when the new `EventConnection` is not initialized within `EventT::Connect` but can be referenced by `EventT::Signal` via `this->connections` map, leading to a segfault. 

Reference from `EventT::Connect`
[...]

PR #3042, opened 2021-07-08
make events threadsafe

lihui815:
Fixes Per issue #2874, race conditions can occur between `EventT::Connect` and `EventT::Signal`. In order to address this, add lock on `Connect`.

However, this only keeps thread-safety between `Connect` (insert) and the `Cleanup` (delete) referenced from `Signal`. There is still potential race condition between the rest of `Signal` (access) and `Connect` (insert). Adding a mutex lock guard here however will lock up the program as there appears to be a `Connect` call from a callback, and using recursive mutexes will impact performance to some extent.

Instead, since the conflict occurs between connection insertion and access, add a null check before accessing to each `Signal` definition.

### DIFF
--- a/gazebo/common/Event.hh
+++ b/gazebo/common/Event.hh
@@ -611,6 +611,7 @@ namespace gazebo
     template<typename T>
     EventT<T>::~EventT()
     {
+      std::lock_guard<std::mutex> lock(this->mutex);
       this->connections.clear();
     }
 
--- a/gazebo/common/Event.hh
+++ b/gazebo/common/Event.hh
@@ -593,7 +593,7 @@ namespace gazebo
       private: EvtConnectionMap connections;
 
       /// \brief A thread lock.
-      private: std::mutex mutex;
+      private: mutable std::mutex mutex;
 
       /// \brief List of connections to remove
       private: std::list<typename EvtConnectionMap::const_iterator>
--- a/gazebo/common/Event.hh
+++ b/gazebo/common/Event.hh
@@ -277,7 +277,7 @@ namespace gazebo
         this->SetSignaled(true);
         for (const auto &iter: this->connections)
         {
-          if (iter.second->on)
+          if ((iter.second != NULL) && iter.second->on)
           {
             IGN_PROFILE_BEGIN("callback0");
             iter.second->callback();
@@ -298,7 +298,7 @@ namespace gazebo
         this->SetSignaled(true);
         for (const auto &iter: this->connections)
         {
-          if (iter.second->on)
+          if ((iter.second != NULL) && iter.second->on)
           {
             IGN_PROFILE_BEGIN("callback1");
             iter.second->callback(_p);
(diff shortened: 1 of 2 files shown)


iche033:
thanks for the fix. Do you have a minimal example that reproduces this segfault?

gazebo/common/Event.hh
```diff
@@ -634,6 +635,7 @@ namespace gazebo
     template<typename T>
     unsigned int EventT<T>::ConnectionCount() const
     {
+      std::lock_guard<std::mutex> lock(this->mutex);
```
j-rivero, line 639: locking `this->mutex` in a `const` method is probably going to make the compiler to cry if we instantiate this. One possible solution (my bet is that ABI checker would be fine) is to use `mutable` for the mutex.
```diff
diff --git a/gazebo/common/Event.hh b/gazebo/common/Event.hh
index 65c47a8beb..79544527a8 100644
--- a/gazebo/common/Event.hh
+++ b/gazebo/common/Event.hh
@@ -593,7 +593,7 @@ namespace gazebo
       private: EvtConnectionMap connections;
 
       /// \brief A thread lock.
-      private: std::mutex mutex;
+      private: mutable std::mutex mutex;
 
       /// \brief List of connections to remove
       private: std::list<typename EvtConnectionMap::const_iterator>
```

I've include it in 14ad930c8d3f82362d28bdee570a6d02531a949e

j-rivero:
PR looks great to me. Do we need to include the mutex lock in the destructor too? I've done it in 0096765a2eaf42091b8b7cb9b2b4cc790c449235. 

I've setup a branch to include changes in destructor, mutable for mutex and an experimental test that plays with 3 threads trying to force a possible race condition: a6e8999cf33a468a9d5a8c62e812ed15b274edaf

These changes are in `jrivero/gazebo11-tse/test` branch in this same repository. If you agree, you can just pull them from my branch. Thanks!

gazebo/common/Event.hh
lihui815: thanks & merged

lihui815:
@j-rivero thanks & merged in PR

j-rivero:
Umm we have a problem on Mac:
```
25: [ RUN      ] EventTest.RaceConditions
25: Start create connections thread
25: Start access thread
25: Start remove thread
25: libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
 25/416 Test  #25: UNIT_Event_TEST .......................................Subprocess aborted***Exception:   3.42 sec
test 26
        Start  26: check_UNIT_Event_TEST

26: Test command: /Users/jenkins/workspace/gazebo-ci-pr_any-homebrew-amd64/gazebo/tools/check_test_ran.py "/Users/jenkins/workspace/gazebo-ci-pr_any-homebrew-amd64/build/test_results/UNIT_Event_TEST.xml"
26: Test timeout computed to be: 10000000
26: Checking for test results in /Users/jenkins/workspace/gazebo-ci-pr_any-homebrew-amd64/build/test_results/UNIT_Event_TEST.xml
26: Cannot find results, writing failure results to /Users/jenkins/workspace/gazebo-ci-pr_any-homebrew-amd64/build/test_results/UNIT_Event_TEST.xml
 26/416 Test  #26: check_UNIT_Event_TEST .................................***Failed    0.06 sec
```
Need to look into it.

j-rivero:
> Need to look into it.

I don't want to block this PR since it improves the current state of things. I've ticketed the issue of the test problem on Mac https://github.com/osrf/gazebo/issues/3063.

PR #3050, opened 2021-07-26
CMake exports: remove -std=c++11 flag

nim65s:
Hello,

Currently, this package exports a `GAZEBO_CXX_FLAGS` containing `-std=c++11`. 
This looks obsolete, as it downgrade the default setting of most compilers for dependent packages.


alexdewar:
I'm the maintainer for the Gazebo package on the AUR and we essentially already do this otherwise it makes it impossible to use Gazebo with C++14 (or higher) code: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=gazebo

So :+1: from me.

scpeters (approved):

chapulina (approved):
Seems reasonable to me

PR #3059, opened 2021-08-10
Move a couple of multilink worlds to to test directory

iche033:
A follow-up PR to #3031.

The `multilink_shape_nested.world` and `multilink_shape_included.world` are used by `ModelListWidget_TEST` and references models in the `test/models` directory so we should just move them to `test/worlds` directory.  They can't be launched in normal simulation mode since gazebo will complain that the models are not found.


j-rivero (approved):
I've restarted the homebrew build to be sure but changes looks good to me. Feel free to merge if CI is happy.

Issue #1949, opened 2016-05-04
Setting GAZEBO_PLUGIN_PATH overrides the compile-time default


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/1949)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
By default, plugins shipped with Gazebo work without the need to set extra environment variables. For example, insert `Submarine (buoyant)` from the insert tab and it floats.

But when you want to use custom plugins and set the `GAZEBO_PLUGIN_PATH` environment variable, for example:

~~~
export GAZEBO_PLUGIN_PATH=`pwd`:$GAZEBO_PLUGIN_PATH
~~~

Then the default plugins stop working. Trying to insert the submarine now prints this message:

`[Err] [Plugin.hh:165] Failed to load plugin libBuoyancyPlugin.so: libBuoyancyPlugin.so: cannot open shared object file: No such file or directory`

This is because, if `GAZEBO_PLUGIN_PATH` is set, we ignore the compile-time path, in [SystemPaths::UpdatePluginPaths()](https://github.com/osrf/gazebo/blob/65e6ec31323b30b3e08b68656b75022a267c4ec7/gazebo/common/SystemPaths.cc#L251)

chapulina:
As a reference, on Ignition we don't let the environment variables prevent the official resources from being found / loaded. So setting `IGN_GAZEBO_RESOURCE_PATH` won't break any core usage, because we always lookup the install paths, and there's no need to source a setup file.

My suggestion to resolve this issue would be to do the same here. That is, always prepend the env var to the compile time paths, instead of overriding them. That would be done here:

https://github.com/osrf/gazebo/blob/17c693050d43c8348e2e64a149ac4f3e6bd93daa/gazebo/common/SystemPaths.cc#L199-L208

PR #3066, opened 2021-08-17
Add colcon.pkg with hook for setup.bash

scpeters:
This will cause gazebo's relocatable setup.bash script (recently added in #3061) to be sourced automatically when sourcing a colcon setup.bash script.

### DIFF
--- a/colcon.pkg
+++ b/colcon.pkg
@@ -1,0 +1,8 @@
+# Configuration file for colcon (https://colcon.readthedocs.io).
+#
+# Please see the doc for the details of the spec:
+#   - https://colcon.readthedocs.io/en/released/user/configuration.html#colcon-pkg-files
+
+{
+    "hooks": ["share/gazebo/setup.bash"]
+}


ivanpauno (approved):

j-rivero:
> This will cause gazebo's relocatable setup.bash script (recently added in #3061) to be sourced automatically when sourcing a colcon setup.bash script.

I have nothing against adding this file but have always seemed weird to me to host "non-core" build tools configurations inside an upstream package like Gazebo. In this case, given the "friendliness" of ROS and Gazebo I think it is fine to add the file but it is difficult to scale if the friends of Conda, vcpkg, or other build tool come to the repo and wanted to add custom files to support any particular behavior.

I think that colcon can handle this nicely by using meta. There is even a file for Gazebo (identical to this one) on it. https://github.com/colcon/colcon-metadata-repository/blob/master/Gazebo.meta

audrow (approved):

scpeters:
> > This will cause gazebo's relocatable setup.bash script (recently added in #3061) to be sourced automatically when sourcing a colcon setup.bash script.
> 
> I have nothing against adding this file but have always seemed weird to me to host "non-core" build tools configurations inside an upstream package like Gazebo. In this case, given the "friendliness" of ROS and Gazebo I think it is fine to add the file but it is difficult to scale if the friends of Conda, vcpkg, or other build tool come to the repo and wanted to add custom files to support any particular behavior.

that's a fair point, and I would consider this not an invitation to host any metadata, but since colcon is one of our supported build tools at Open Robotics, that it doesn't hurt to host the file

> I think that colcon can handle this nicely by using meta. There is even a file for Gazebo (identical to this one) on it. https://github.com/colcon/colcon-metadata-repository/blob/master/Gazebo.meta

ooh, that's a good point, though I've never configured that metadata repository before. I'll solicit a few more opinions before making a decision about merging

chapulina:
I lean towards the same opinion as @j-rivero , this is setting a precedent that's difficult to scale, and colcon provides a way of handling this without hardcoding the file into this repository.

Also note that this would be overriding this bit of logic which is meant to take the compile-time default paths unless the user explicitly sets the environment variables:

https://github.com/osrf/gazebo/blob/17c693050d43c8348e2e64a149ac4f3e6bd93daa/gazebo/common/SystemPaths.cc#L199-L208

More info on https://github.com/osrf/gazebo/issues/1949

scpeters:
> I lean towards the same opinion as @j-rivero , this is setting a precedent that's difficult to scale, and colcon provides a way of handling this without hardcoding the file into this repository.

understood

> Also note that this would be overriding this bit of logic which is meant to take the compile-time default paths unless the user explicitly sets the environment variables:

I understand that logic to be a backstop in case the users don't source the `setup.sh` file or create the needed environment variables. I don't see how automatically sourcing the file is a problem.

One other small point: I realized while working on #3061 that our compile-time default paths are not relocatable. If the installed files are relocated, then sourcing the new `setup.bash` is the only way to properly configure the paths. For now at least, the centrally hosted [colcon meta file](https://github.com/colcon/colcon-metadata-repository/blob/master/Gazebo.meta) adds a hook for `setup.sh`, which is not relocatable. It's not so easy to propose changing the meta file to use `setup.bash` though because there are so many released binaries of gazebo without `setup.bash`.

In the end, this pull request isn't necessary, so I'm leaning towards declining it since there isn't consensus.

PR #3069, opened 2021-08-20
Load both .so and .dylib plugins on macOS

### LINKED ISSUE #800: Plugin loading on OS X

scpeters:
As discussed in #800, macOS supports dynamic libraries with extensions `.so` and `.dylib`, but `gazebo::common::Plugin::Create` only loads `.dylib` files on macOS. This refactors that function to try loading both `.so` and `.dylib` files on macOS.

This will fix some test failures on macOS, since the test plugins are [created as `MODULE` libraries](https://github.com/osrf/gazebo/blob/gazebo11/test/plugins/CMakeLists.txt#L34) in cmake, which have extension `.so` on macOS.

* https://build.osrfoundation.org/job/gazebo-ci-gazebo11-homebrew-amd64/85/testReport/(root)/PluginTest/ModelExceptionConstructor/
* https://build.osrfoundation.org/job/gazebo-ci-gazebo11-homebrew-amd64/85/testReport/(root)/PluginTest/ModelExceptionInit/
* https://build.osrfoundation.org/job/gazebo-ci-gazebo11-homebrew-amd64/85/testReport/(root)/PluginTest/ModelExceptionLoad/

~~~
469: [0m[1;32m[0m[1;32m[0m[1;31m[Err] [Plugin.hh:178] [0m[1;31mFailed to load plugin [0m[1;31mlibExceptionModelPluginConstructor.dylib[0m[1;31m: [0m[1;31mdlopen(libExceptionModelPluginConstructor.dylib, 9): image not found[0m[1;31m
~~~

Fixes #800.


scpeters:
I'm not sure we have any plugin loading tests running on Windows. Someone should test this explicitly on Windows with plugins to ensure it doesn't break anything.

scpeters:
> I'm not sure we have any plugin loading tests running on Windows. Someone should test this explicitly on Windows with plugins to ensure it doesn't break anything.

I haven't seen any volunteers to test on windows, so I think we can proceed with code review and our CI results on Ubuntu and macOS. The diff is a little messy because code is moved around in the file, but it currently shows no changes to the windows `ifdef` block

j-rivero (approved):
Code changes and reasoning looks good to me. A part of looking a couple of improvements:
 * Stop messing with non portable path strings and boost::path and replace it all with std::filesystem::path (probably supported in all platforms since we are using c++17).
 * Report to the user all paths and filenames used to check for the plugin when the search fails.

They are probably out of the scope here anyway.

I don't see in the changes why Windows should fail.

scpeters:
> Code changes and reasoning looks good to me. A part of looking a couple of improvements:
> 
> * Stop messing with non portable path strings and boost::path and replace it all with std::filesystem::path (probably supported in all platforms since we are using c++17).

I agree that we could consider this, but as it seems like a nice-to-have, we can save it for another day

> * Report to the user all paths and filenames used to check for the plugin when the search fails.

 We currently check all the paths in `GAZEBO_PLUGIN_PATH` for the specified file. If no matching file is found in those paths, we try to `dlopen` the file without a path, which should search `LD_LIBRARY_PATH`. If no file is found, we print an error that looks like the following:

~~~
[Err] [Plugin.hh:212] Failed to load plugin libHydraPlugin.dylib: dlopen(libHydraPlugin.dylib, 9): image not found
~~~

I suppose we could add an extra message if the plugin was not found in `GAZEBO_PLUGIN_PATH` and that `LD_LIBRARY_PATH` is being searched instead

j-rivero:
> > *
> 
> We currently check all the paths in `GAZEBO_PLUGIN_PATH` for the specified file. If no matching file is found in those paths, we try to `dlopen` the file without a path, which should search `LD_LIBRARY_PATH`. If no file is found, we print an error that looks like the following:

Yep, I was thinking on an improved error message, something like:

```
[Err] [Plugin.hh:212] Failed to load plugin libHydraPlugin.dylib
[Err] [Plugin.hh:212]   - paths used to found the plugin in GAZEBO_PLUGIN_PATH: path1, path2, path3, ... 
[Err] [Plugin.hh:212]   - also tried to found it using LD_LIBRARY_PATH with the name of libHydraPlugin.dylib
[Err] [Plugin.hh:212]   - also tried to found the plugin with the name libHydraPlugin.so
```
Don't worry, it is work for a different PR.

scpeters:
> > > 
> > 
> > 
> > We currently check all the paths in `GAZEBO_PLUGIN_PATH` for the specified file. If no matching file is found in those paths, we try to `dlopen` the file without a path, which should search `LD_LIBRARY_PATH`. If no file is found, we print an error that looks like the following:
> 
> Yep, I was thinking on an improved error message, something like:
> 
> ```
> [Err] [Plugin.hh:212] Failed to load plugin libHydraPlugin.dylib
> [Err] [Plugin.hh:212]   - paths used to found the plugin in GAZEBO_PLUGIN_PATH: path1, path2, path3, ... 
> [Err] [Plugin.hh:212]   - also tried to found it using LD_LIBRARY_PATH with the name of libHydraPlugin.dylib
> [Err] [Plugin.hh:212]   - also tried to found the plugin with the name libHydraPlugin.so
> ```
> 
> Don't worry, it is work for a different PR.

ok, thanks then I'll go ahead and merge this as is

PR #3074, opened 2021-08-26
👩‍🌾 Remove bitbucket-pipelines

### LINKED ISSUE #203: Dependency on bitbucket-pipelines.yml should be removed
Several jobs rely on `bitbucket-pipelines.yml` to find out what Ubuntu distribution to use when running CI, for example:

https://github.com/ignition-tooling/release-tools/blob/8e8f5e4d2858de04643e666abd07772b156f98e6/jenkins-scripts/dsl/ignition.dsl#L241

Now that BitBucket is dead, we should get this information from somewhere else and remove those files from all projects.

chapulina:
Fixes https://github.com/ignition-tooling/release-tools/issues/203

This will make CI default to Bionic, so nothing would change.

Targeting `gazebo11` because the `gazebo9` branch doesn't have a `bitbucket-pipelines.yml` already.

---

https://github.com/osrf/buildfarmer/issues/224

### DIFF
--- a/bitbucket-pipelines.yml
+++ b/bitbucket-pipelines.yml
@@ -1,7 +1,0 @@
-image: ubuntu:bionic
-
-pipelines:
-  default:
-    - step:
-        script:
-          - apt-get update


chapulina:
The Ubuntu job, which was the one using this file, built fine.

j-rivero (approved):

PR #3076, opened 2021-08-30
Fix pkgconfig prefix with absolute CMAKE_INSTALL_LIBDIR

lopsided98:
#2879 introduced an assumption that `CMAKE_INSTALL_LIBDIR` is relative by concatenating it with `CMAKE_INSTALL_PREFIX`. Instead, this path should be formed using `CMAKE_INSTALL_FULL_LIBDIR`.

cc @Pro @scpeters


scpeters (approved):

scpeters:
thanks for the fix!

PR #3116, opened 2021-10-07
Always publish if flag is passed to Model::SetScale 

scpeters:
Currently `Model::SetScale` does nothing if the new scale is approximately equal to the current scale, even if the `_publish` flag is set to true. This change ensures that the scale is always published if the flag is set.


scpeters:
cc @mogumbo

scpeters:
diff without whitespace: https://github.com/osrf/gazebo/pull/3116/files?w=1

iche033 (approved):
looks good to me.

PR #3121, opened 2021-10-14
Fix gzclient starting with black screen

scpeters:
There is a longstanding issue (#681) in which `gzclient` occasionally starts with a black screen, failing to ever display the scene. Opening a second `gzclient` instance often works properly, suggesting a race condition in initialization. I highly suspect much of the non-determinism comes from the use of the `~/request` topic with gazebo-transport to communicate `scene_info` from `gazebo::physics::World` to `gazebo::rendering::Scene`. It is more reliable to use an ignition-transport service, so I have added a `/scene_info` ignition-transport service to `World` and changed `Scene` to prefer calling this service instead of using the `~/request` topic. If the ignition service is unavailable or fails, it will fall back to the current method using `~/request`, which should preserve backwards compatibility between old and new instances of `gzserver` and `gzclient`.

I've added a test for the new service to `INTEGRATION_info_services`. I've also fixed a few compiler warnings in ab57017.

### DIFF
--- a/gazebo/rendering/Scene.cc
+++ b/gazebo/rendering/Scene.cc
@@ -454,7 +454,6 @@ void Scene::Init()
 
   // Get scene info from physics::World with ignition transport service
   ignition::transport::Node node;
-  ignition::msgs::Scene sceneInfo;
   const std::string serviceName = "/scene_info";
   std::vector<ignition::transport::ServicePublisher> publishers;
   if (!node.ServiceInfo(serviceName, publishers) ||
--- a/gazebo/rendering/Scene.cc
+++ b/gazebo/rendering/Scene.cc
@@ -430,9 +430,21 @@ void Scene::Init()
       this->dataPtr->worldVisual));
   this->dataPtr->originVisual->Load();
 
-  this->dataPtr->requestPub->WaitForConnection();
-  this->dataPtr->requestMsg = msgs::CreateRequest("scene_info");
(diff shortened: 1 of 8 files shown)


WilliamLewww:
There is a small merge conflict but other than that, the changes seem to be working!

scpeters:
> There is a small merge conflict but other than that, the changes seem to be working!

I just merged with `gazebo11`, conflicts are resolved

WilliamLewww (approved):

gazebo/rendering/Scene.cc
```diff
@@ -454,9 +452,21 @@ void Scene::Init()
       this->dataPtr->worldVisual));
   this->dataPtr->originVisual->Load();
 
-  this->dataPtr->requestPub->WaitForConnection();
-  this->dataPtr->requestMsg = msgs::CreateRequest("scene_info");
-  this->dataPtr->requestPub->Publish(*this->dataPtr->requestMsg);
+  // Get scene info from physics::World with ignition transport service
+  ignition::transport::Node node;
+  ignition::msgs::Scene sceneInfo;
```
scpeters: I just realized this variable is unused. I'll wait for the current CI run to finish, then make this change and merge without CI

scpeters: https://github.com/osrf/gazebo/pull/3121/commits/37556f27f3e57b59e79b4af8e77ecf76e0ad8499

Issue #3125, opened 2021-10-22
Slow-loading sensor plugin causes failure to load model plugins


scpeters:
Since https://github.com/osrf/gazebo/pull/3121 was recently merged, I have noticed a problem with some models that are slow to load. Specifically, if a model is spawned that contains a sensor plugin that is slow to load, it is possible that model plugins will fail to load, with the following error message:

~~~
[Err] [Model.cc:1061] Sensors failed to initialize when loading model[...] via the factory mechanism. Plugins for the model will not be loaded.
~~~

I believe this is caused by [this logic](https://github.com/osrf/gazebo/blob/gazebo11_11.8.1/gazebo/physics/Model.cc#L1025-L1036) in `Model::LoadPlugins` that skips loading of model plugins if the sensors take longer than 5 seconds to initialize, which includes the time required to load sensor plugins.

I have created a world file along with two test plugins to reproduce the failure in https://github.com/scpeters/gazebo/commit/8a2a4cca39beb055f7236a28ecbac113f567dd6c. The world has a `WorldSpawnModelPlugin` that spawns a model that has a single box with a camera sensor using the `SlowLoadingSensorPlugin` and the `InitialVelocityPlugin` as a model plugin. The box should be given an initial velocity after loading, but the slow-loading sensor plugin causes the model plugin to not be loaded, so the box remains at its initial pose.

scpeters:
Closed by #3126

PR #3126, opened 2021-10-22
Fix slow loading of spawned model with plugins

scpeters:
As reported in #3125, since #3121 was recently merged I have noticed a problem with some models that are slow to load. Specifically, if a model is spawned that contains a sensor plugin that is slow to load, it is possible that model plugins will fail to load, with the following error message:

~~~
[Err] [Model.cc:1061] Sensors failed to initialize when loading model[...] via the factory mechanism. Plugins for the model will not be loaded.
~~~

I believe this is caused by [this logic](https://github.com/osrf/gazebo/blob/gazebo11_11.8.1/gazebo/physics/Model.cc#L1025-L1036) in `Model::LoadPlugins` that skips loading of model plugins if the sensors take longer than 5 seconds to initialize, which includes the time required to load sensor plugins. I have increased this timeout to 50 seconds in https://github.com/osrf/gazebo/commit/2589d13b05919d58102ebb0ae5e8c435fa2ba57d, though I am open to suggestions for how to make this a configurable parameter.

I have added a regression test that loads a world file along with two test plugins to reproduce the failure. The world has a `WorldSpawnModelPlugin` that spawns a model that has a single box with a camera sensor using the `SlowLoadingSensorPlugin` and the `InitialVelocityPlugin` as a model plugin. The box should be given an initial velocity after loading.

### DIFF
--- a/gazebo/physics/Model.cc
+++ b/gazebo/physics/Model.cc
@@ -1034,11 +1034,12 @@ void Model::LoadPlugins()
   if (this->GetPluginCount() > 0)
   {
     int iterations = 0;
+    const int maxIterations = 500;
 
     // Wait for the sensors to be initialized before loading
     // plugins, if there are any sensors
     while (this->GetSensorCount() > 0 && !this->world->SensorsInitialized() &&
-           iterations < 50)
+           iterations < maxIterations)
     {
       common::Time::MSleep(100);
       iterations++;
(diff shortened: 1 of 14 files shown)


scpeters:
> I am open to suggestions for how to make this a configurable parameter.


I suppose we could add an SDFormat world parameter?

scpeters:
> I am open to suggestions for how to make this a configurable parameter.

I've added an SDFormat parameter in  the ignition xml namespace in https://github.com/osrf/gazebo/pull/3126/commits/35fb92125a9a11680cad489d8a05106e485667f7 and restored the default timeout to 5 seconds

jacobperron (approved):
The changes LGTM.

I would consider increasing the default timeout if you think users are more likely to run into the issue after updating Gazebo.

jacobperron (approved):

PR #3127, opened 2021-10-25
GUI option to change render rate

WilliamLewww:
The effect is not easily seen due to the choppiness of the GIF, but the render rate switches between 62 -> 15.

render_options


gazebo/gui/RenderOptionsWindowPrivate.hh
```diff
@@ -0,0 +1,42 @@
...
+  namespace gui
+  {
+    /// \class RenderOptionsWindowPrivate RenderOptionsWindowPrivate.hh
+    /// \brief Private data for the RenderOptionsWindowPrivate class.
+    class RenderOptionsWindowPrivate
+    {
+      /// \brief Button used to finalize render options.
+      public: QPushButton *okayButton;
+
+      /// \brief QT widget for reading the render rate used in the cloned
```
scpeters: whitespace at end of line

gazebo/gui/RenderOptionsWindow.hh
```diff
@@ -0,0 +1,70 @@
...
+{
+  namespace gui
+  {
+    class RenderOptionsWindowPrivate;
+
+    /// \addtogroup gazebo_gui GUI
+    /// \brief The graphical user interface
+    /// \{
+
+    /// \class RenderOptionsWindow RenderOptionsWindow.hh
```
scpeters: whitespace at end of line

gazebo/gui/RenderOptionsWindowPrivate.hh
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef _GAZEBO_CLONE_WINDOW_PRIVATE_HH_
+#define _GAZEBO_CLONE_WINDOW_PRIVATE_HH_
```
scpeters: fix header guard

gazebo/rendering/UserCamera.hh
```diff
@@ -294,6 +294,12 @@ namespace gazebo
       /// \param[in] _msg New pose message.
       private: void OnJoyPose(ConstPosePtr &_msg);
 
+      /// \brief Handles incoming relative joystick messages.
+      /// Incoming joystick messages are used to control
+      /// translation and rotation rates of the camera position.
+      /// \param[in] _msg New joystick message.
```
scpeters: replace `joystick` in these doc-strings with proper description

scpeters (commented):
merging with gazebo11 should fix the ABI checker

also, I think you could add the render rate to the GUI section of the ModelListWidget on the left-hand side of the screen by adding some code to [ModelListWidget::FillUserCamera](https://github.com/osrf/gazebo/blob/gazebo11_11.8.1/gazebo/gui/ModelListWidget.cc#L3421), since that method already has a pointer to the user camera. If you do this, then you could remove the new code for RenderOptionsWindow, so you can ignore any comments I made about those files if you end up removing the code

WilliamLewww:
> merging with gazebo11 should fix the ABI checker
> 
> also, I think you could add the render rate to the GUI section of the ModelListWidget on the left-hand side of the screen by adding some code to [ModelListWidget::FillUserCamera](https://github.com/osrf/gazebo/blob/gazebo11_11.8.1/gazebo/gui/ModelListWidget.cc#L3421), since that method already has a pointer to the user camera. If you do this, then you could remove the new code for RenderOptionsWindow, so you can ignore any comments I made about those files if you end up removing the code

Changed the position of the render rate tool.
gui

scpeters:
nice work! I noticed a test failure and weird console error message while testing, so I've fixed those in https://github.com/osrf/gazebo/pull/3127/commits/a6a827d92cbd407a149337e9c971b031cafd1eb1

gazebo/gui/ModelListWidget.hh
```diff
@@ -24,6 +24,7 @@
 #include <ignition/msgs/plugin_v.pb.h>
 
 #include "gazebo/gui/qt.h"
+#include "gazebo/gui/RenderWidget.hh"
```
scpeters: I think we could use a forward declaration a few lines below as `class RenderWidget;` instead of including this header file

gazebo/gui/ModelListWidget.hh
```diff
@@ -44,6 +45,8 @@ namespace gazebo
       public: explicit ModelListWidget(QWidget *_parent = 0);
       public: virtual ~ModelListWidget();
 
+      public: void ConnectRenderWidget(RenderWidget* renderWidget);
```
scpeters: add doc-string for the new function, even though most others don't have them

gazebo/gui/ModelListWidgetPrivate.hh
```diff
@@ -132,6 +133,8 @@ namespace gazebo
 
       /// \brief Node for ignition transport communication.
       public: ignition::transport::Node ignNode;
+
+      public: RenderWidget* renderWidget;
```
scpeters: add doc-string for the new member variable, even though some others don't have them

gazebo/gui/ModelListWidget.cc
```diff
@@ -843,10 +851,27 @@ void ModelListWidget::GUICameraPropertyChanged(QtProperty *_item)
...
+  QtProperty *cameraRenderRateProperty = this->ChildItem(cameraProperty,
+      "render rate");
+  if (cameraRenderRateProperty && changedProperty == "render rate")
+  {
+    rendering::UserCameraPtr cam = gui::get_active_camera();
+
+    if (cam)
+    {
+      cam->SetRenderRate(this->dataPtr->variantManager->value(
+          this->ChildItem(cameraRenderRateProperty, "render rate")).toDouble());
```
scpeters: nit: this value is duplicated; instead use a variable to store this value and reference it on this line and in the statement below

~~~
double renderRate = this->dataPtr->variantManager->value(
    this->ChildItem(cameraRenderRateProperty, "render rate")).toDouble();
~~~

scpeters (commented):
I've noticed a few more things; I will fix them myself in another commit

gazebo/gui/ModelListWidget.cc
scpeters: https://github.com/osrf/gazebo/pull/3127/commits/ef876fda10031a6dc199aff4953fa155d31b6390

gazebo/gui/ModelListWidgetPrivate.hh
scpeters: https://github.com/osrf/gazebo/pull/3127/commits/ef876fda10031a6dc199aff4953fa155d31b6390

gazebo/gui/ModelListWidget.hh
scpeters: https://github.com/osrf/gazebo/pull/3127/commits/ef876fda10031a6dc199aff4953fa155d31b6390

gazebo/gui/ModelListWidget.hh
scpeters: https://github.com/osrf/gazebo/pull/3127/commits/ef876fda10031a6dc199aff4953fa155d31b6390

scpeters:
Mark just asked how feasible it would be to specify the GUI render rate from the `<gui>` element in a world file, and as I started thinking about how to implement it, it occurs to me that perhaps we should create a GUI event for changing the render rate instead of adding the `ModelListWidget::ConnectRenderWidget` method.

what do you think @iche033 ?

scpeters:
> Mark just asked how feasible it would be to specify the GUI render rate from the `<gui>` element in a world file, and as I started thinking about how to implement it, it occurs to me that perhaps we should create a GUI event for changing the render rate instead of adding the `ModelListWidget::ConnectRenderWidget` method.
> 
> what do you think @iche033 ?

I've implemented the approach using a GUI event in https://github.com/osrf/gazebo/pull/3127/commits/595f23581a94eef2f74fbd5235f53ef52751ee98

iche033:
> I've implemented the approach using a GUI event in 595f235

yep, looks good to me.

scpeters:
I used `invokeMethod` in https://github.com/osrf/gazebo/pull/3127/commits/e9e162ed96fb28f80af42c74d3b6d2ceab009477 at @iche033's suggestion to address some threading issues I found while testing an approach to loading the render rate from a world file

scpeters (approved):
I approve, though I also made a few changes to it

scpeters:
I'm not going to try reading the render rate from a world file in this PR

iche033 (approved):
works for me.

gazebo/gui/ModelListWidget.cc
HighPriest: Shouldn't the property name be set to `render_rate` as everything else in these menus is using this type of notation?
@scpeters @WilliamLewww

HighPriest:
Is usage of this option documented somewhere? How can it be updated, other than manually in the GUI?
Can this property be set in a world file?
e.g
```xml
<gui>
  <camera>
    <render_rate>30</render_rate>
  </camera>
</gui>
```

It would be very useful, when running on a target machine, to reduce gzclient impact on performance.

scpeters:
> Is usage of this option documented somewhere? How can it be updated, other than manually in the GUI? Can this property be set in a world file? e.g
> 
> ```
> <gui>
>   <camera>
>     <render_rate>30</render_rate>
>   </camera>
> </gui>
> ```
> 
> It would be very useful, when running on a target machine, to reduce gzclient impact on performance.

there is very simple code to set render rate from SDFormat, but it requires a change to the `gui_camera` protobuf message, which breaks ABI, so it can't be released in gazebo11 (see https://github.com/scpeters/gazebo/commit/07ceabf41a065c69205da0e05c7103c1bfb5461a).

an alternative approach could be made that creates an ignition-transport service to allow `gzclient` to query `gzserver` for the render_rate value supplied in the SDFormat world file, similar to the approach in https://github.com/osrf/gazebo/commit/42f7e48c8731163e6223125dfde3c811c9d92a7f. This was more complex, and I did not have time to implement this when working on this pull request

PR #3135, opened 2021-11-23
Wideanglecamera equidistant mapping bounding fix

kjeppesen1:
Currently when using a `wideanglecamera` sensor with a lens type of `equidistant`, the actual image data gets restricted to a small circular region with the rest of the image having a gray ring around it. See the images below for the image that is produced for an `equidistant` lens with an HFOV of 1.5708, 2.0, and 3.14159.
equidistant_1p5708 
equidistant_2p0
equidistant_3p14159

The issue appears to be in `wide_lens_map_fp.glsl` on Line 38. The value `theta` gets calculated with `asin(param)` driving the equation if `fun.x` is 1 (which is true for `equisolid_angle` and `orthographic` lens types). The way this equation is written becomes problematic for other lens types, though, when the `param > 1` since the resulting value of `asin(param)` becomes undefined.

This PR is to add a check for function type to avoid the undefined `asin` computation. See the images below for an `equidistant` lens with an HFOV of 1.5708, 2.0, and 3.14159 after this fix.
equidistant_1p5708_fixed
equidistant_2p0_fixed
equidistant_3p14159_fixed

### DIFF
--- a/media/materials/scripts/gazebo.material
+++ b/media/materials/scripts/gazebo.material
@@ -1515,7 +1515,7 @@ fragment_program Gazebo/WideLensMapFS glsl
     param_named c2 float 1
     param_named c3 float 0
     param_named f float 1
-    param_named fun float3 0 1
+    param_named fun float3 0 0 1
     param_named cutOffAngle float 3.14
   }
 }
--- a/media/materials/programs/wide_lens_map_fp.glsl
+++ b/media/materials/programs/wide_lens_map_fp.glsl
@@ -35,7 +35,13 @@ void main()
 {
   // calculate angle from optical axis based on the mapping function specified
   float param = r/(c1*f);
-  float theta = fun.x*asin(param)+fun.y*atan(param)+fun.z*param;
+  float theta = 0.0;
+  if(fun.x > 0)
+    theta = asin(param);
+  else if(fun.y > 0)
+    theta = atan(param);
+  else if(fun.z > 0)
+    theta = param;
   theta = (theta-c3)*c2;
 
   // compute the direction vector that will be used to sample from the cubemap


kjeppesen1:
@scpeters @iche033

media/materials/scripts/gazebo.material
```diff
@@ -1515,7 +1515,7 @@ fragment_program Gazebo/WideLensMapFS glsl
     param_named c2 float 1
     param_named c3 float 0
     param_named f float 1
-    param_named fun float3 0 1
+    param_named fun float3 0 0 1
```
kjeppesen1, line 1518: This line might not change anything, but it seemed odd to not complete the default float3 values

mogumbo: If `float3 0 1` was working before, wouldn't Ogre have been interpreting it as `float3 0 1 0`? This looks like it would change behavior from using `atan(param)` to using `param`.

scpeters: I agree, this is suspicious

iche033: I think it could be `0 0 1` to match the default `fun` value of `id` specified [here](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/rendering/CameraLensPrivate.hh#L141) (which is defined to be `UnitZ` [here](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/rendering/CameraLensPrivate.hh#L76).) 

It may not matter that much though as the values specified here in the material script are the initial values but should be overriden every frame by the [SetUniformVariables call](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/rendering/WideAngleCamera.cc#L383-L384)

iche033 (approved):
good catch! looks good to me

media/materials/scripts/gazebo.material
scpeters: 👍

scpeters (approved):

PR #3060, opened 2021-08-11
Distortion initialization fix with background color

WilliamLewww:
Closes #2527 - Last camera's distortion affects previous cameras' distortions

When setting the background color to a scene, the distortion parameters continue to leak. From #3033 we can infer that setting the scene's background color is effecting the distortion compositor which is causing the improper order for compositor initialization.

To fix the issue, we reinitialize the distortion after setting the scene's color to ensure the distortion compositor is properly set for each camera.

PR #3033, opened 2021-06-25
Distortion camera initialization fix

WilliamLewww:
Closes #2527  - Last camera's distortion affects previous cameras' distortions

When loading multiple cameras on the application startup, each camera would finish initialization but their compositor's material would only be set after every camera has been initialized. This would make all compositor materials to be set to the most recently initialized camera.

To fix this issue, `the compositor material should be set right after their camera is done initializing` rather than `waiting for all the cameras to finish initialization`.

fix

PR #3136, opened 2021-11-24
Workaround for wrong distortion material applied to some cameras

mogumbo:
When using stereo cameras with different distortion parameters and attaching other Ogre compositors to the cameras, the same distortion parameters will be applied to both cameras. This PR implements a workaround.

It looks to me like Distortion.cc tries to do the right thing by using a unique Ogre::CompositorInstance for each camera. Unfortunately, it looks like these CompositorInstances share the same CompositionTechnique inside Ogre, causing passes and materials to get stomped from one CompositorInstance to the next. This stomping only happens if you add compositors to your cameras in addition to the distortion compositor.

### DIFF
--- a/gazebo/rendering/Distortion.cc
+++ b/gazebo/rendering/Distortion.cc
@@ -74,6 +74,9 @@ namespace gazebo
       /// \brief Ogre Material that contains the distortion shader
       public: Ogre::MaterialPtr distortionMaterial;
 
+      /// \brief Ogre Texture that contains the distortion map
+      public: Ogre::TexturePtr distortionTexture;
+
       /// \brief Connection for the pre render event.
       public: event::ConnectionPtr preRenderConnection;
 
@@ -91,6 +94,14 @@ namespace gazebo
       virtual void notifyMaterialRender(Ogre::uint32 _passId,
                                         Ogre::MaterialPtr& _material)
       {
+        // If more compositors are added to the camera in addition to the
+        // distortion compositor, an Ogre bug will cause the material for the
+        // last camera initialized to be used for all cameras. This workaround
+        // doesn't correct the material used, but it applies the correct texture
+        // to this active material.
+        _material->getTechnique(0)->getPass(_passId)->getTextureUnitState(1)->
+          setTexture(distortionTexture);
+      
         // @todo Explore more efficent implementations as it is run every frame
         Ogre::GpuProgramParametersSharedPtr params =
             _material->getTechnique(0)->getPass(_passId)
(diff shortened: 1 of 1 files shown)


gazebo/rendering/Distortion.cc
```diff
@@ -91,6 +94,14 @@ namespace gazebo
...
                                         Ogre::MaterialPtr& _material)
       {
+        // If more compositors are added to the camera in addition to the
+        // distortion compositor, an Ogre bug will cause the material for the
+        // last camera initialized to be used for all cameras. This workaround
+        // doesn't correct the material used, but it applies the correct texture
+        // to this active material.
+        _material->getTechnique(0)->getPass(_passId)->getTextureUnitState(1)->
+          setTexture(distortionTexture);
+
```
scpeters: nit: I believe there is trailing whitespace on this line

scpeters:
this fixes the issues I'm seeing in the VIPER simulation. I've even tried reverting the fixes from #3033 and #3060 with the following patch and everything still seems to work

~~~
diff --git a/gazebo/rendering/Camera.cc b/gazebo/rendering/Camera.cc
index 502f122a13..819b2a5a44 100644
--- a/gazebo/rendering/Camera.cc
+++ b/gazebo/rendering/Camera.cc
@@ -1630,7 +1630,6 @@ void Camera::SetRenderTarget(Ogre::RenderTarget *_target)
     if (this->dataPtr->distortion)
     {
       this->dataPtr->distortion->SetCamera(shared_from_this());
-      this->renderTarget->update();
     }
 
     if (this->GetScene()->GetSkyX() != NULL)
@@ -2111,15 +2110,6 @@ bool Camera::SetBackgroundColor(const ignition::math::Color &_color)
   if (this->OgreViewport())
   {
     this->OgreViewport()->setBackgroundColour(Conversions::Convert(_color));
-
-    // refresh distortion to prevent improper compositor initialization
-    // https://github.com/osrf/gazebo/pull/3033
-    if (this->dataPtr->distortion)
-    {
-      this->dataPtr->distortion->RefreshCompositor(shared_from_this());
-      this->renderTarget->update();
-    }
-
     return true;
   }
   return false;
~~~

scpeters:
this appears to be a more complete fix for #2527

mogumbo:
The new title "Fix Camera Distortion initialization: use correct material" is a bit misleading. My workaround actually leaves the incorrect material in place but updates its distortion texture each time a different camera renders.

scpeters:
> The new title "Fix Camera Distortion initialization: use correct material" is a bit misleading. My workaround actually leaves the incorrect material in place but updates its distortion texture each time a different camera renders.

thanks; I've restored your original title

iche033:
ok yeah it does look like the compositor instances share the same compositor technique inside ogre. Changes look good to me.

scpeters:
homebrew CI is failing due to https://github.com/ignition-tooling/release-tools/issues/573, but otherwise the CI results look fine to me

gazebo/rendering/Distortion.cc
mogumbo: fixed

scpeters:
rerunning macOS CI

scpeters (approved):

PR #3138, opened 2021-11-27
Fix setup.sh and install dirs with absolute paths

lopsided98:
#3061 introduced a similar problem to the one I fixed in #3076. `CMAKE_INSTALL_*DIR` can be absolute, so it cannot be concatenated with `CMAKE_INSTALL_PREFIX`. For `setup.sh` though, we still want to use relative paths if possible to keep it relocatable. I implemented this similarly to the pkgconfig file, using the `join_paths` function. With this patch, if the `CMAKE_INSTALL_*DIR` variables are relative, `setup.sh` uses paths relative to `${installPrefix}` (as before), whereas if those variables are absolute, `setup.sh` contains absolute paths. The `GAZEBO_*_PATH` variables are made always absolute using `CMAKE_INSTALL_FULL_*DIR`, retaining the existing behavior. I haven't tested this on Windows, so you may want to verify that `setup.bat` still contains the correct values.

cc @scpeters


scpeters (approved):
thanks for the fix! Hopefully I'll remember next time how to do this properly so you won't need to fix it so frequently

PR #3140, opened 2021-12-02
setup.sh: fix relocatability of LD_LIBRARY_PATH

scpeters:
The LD_LIBRARY_PATH variable is currently set from the non-relocatable `@GAZEBO_PLUGIN_PATH@` configure-time value. It can be made relocatable by evaluating the `$GAZEBO_PLUGIN_PATH` environment variable that is set earlier in `setup.sh`.

cc @lopsided98 since you have touched this file recently.


jacobperron (approved):
LGTM

PR #3154, opened 2022-01-05
Load model plugins even after sensor timeout

scpeters:
Currently model plugins are not loaded in certain cases if it takes too long to initialize sensors. In #3126, we made the timeout configurable and added a test. This pull request changes the behavior to print an error message if the sensors fail to initialize before the timeout but to load the model plugins anyway.

The test is changed to use a very short timeout, so that we can confirm see the error message and confirm that the model plugins are still loaded.

### DIFF
--- a/gazebo/physics/Model.cc
+++ b/gazebo/physics/Model.cc
@@ -1051,9 +1051,18 @@ void Model::LoadPlugins(unsigned int _timeout)
       iterations++;
     }
 
-    // Load the plugins if the sensors have been loaded, or if there
-    // are no sensors attached to the model.
-    if (iterations < maxIterations)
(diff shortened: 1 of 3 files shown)


adityapande-1995:
Building and testing locally...

scpeters:
CI looks ok to me. The 3125 regression test passes:

* https://build.osrfoundation.org/job/gazebo-ci-pr_any-ubuntu_auto-amd64-gpu-nvidia/861/testReport/(root)/Issue3125Test/
* https://build.osrfoundation.org/job/gazebo-ci-pr_any-homebrew-amd64/2629/testReport/(root)/Issue3125Test/

adityapande-1995 (approved):
Looks good ! Homebrew failures seem to be unrelated

PR #3160, opened 2022-01-14
Revert boost/bind.hpp changes in header files and bump to 11.10.1

scpeters:
As noted in https://github.com/osrf/gazebo/pull/3156#issuecomment-1012404375, the changes in #3156 to header files have the potential to break downstream software. This pull request reverts the changes to header files and bumps the patch version to 11.10.1, so we can make a quick release and unbreak downstream software.

cc @traversaro (apologies to @alexdewar )

Update: I started wondering if this would affect gazebo_ros_pkgs, since I see `_1` in [gazebo_ros_api_plugin on `noetic-devel`](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/2.9.2/gazebo_ros/src/gazebo_ros_api_plugin.cpp#L158), but it didn't fail its latest CI job (using Ubuntu 20.04):

* [Build Status](https://build.osrfoundation.org/job/ros_gazebo_pkgs-ci-default_noetic-focal-amd64/24/) https://build.osrfoundation.org/job/ros_gazebo_pkgs-ci-default_noetic-focal-amd64/24/

I believe this is because `boost/thread.hpp` provides an opportunistic include of `boost/bind.hpp` for boost 1.72 and earlier (see https://github.com/boostorg/thread/compare/boost-1.72.0...boost-1.73.0#diff-c94867ed785b09487490187250e85510d4ac8698e7fa22dc56906baeb83423d0L33), so this does not affect Ubuntu 20.04, which uses boost 1.71. That's also why the homebrew CI failed in #3156 but not the Ubuntu CI.

So I still support the changes in this pull request, but I've convinced myself that it is not as dire as it sounded, since it does not affect Ubuntu 20.04.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ set (GAZEBO_MINOR_VERSION 10)
 # The patch version may have been bumped for prerelease purposes; be sure to
 # check gazebo-release/ubuntu/debian/changelog@default to determine what the
 # next patch version should be for a regular release.
-set (GAZEBO_PATCH_VERSION 0)
+set (GAZEBO_PATCH_VERSION 1)
 
 set (GAZEBO_VERSION ${GAZEBO_MAJOR_VERSION}.${GAZEBO_MINOR_VERSION})
 set (GAZEBO_VERSION_FULL ${GAZEBO_MAJOR_VERSION}.${GAZEBO_MINOR_VERSION}.${GAZEBO_PATCH_VERSION})
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,9 @@
 1. Revert boost/bind.hpp changes in header files
     * [Pull request #3160](https://github.com/osrf/gazebo/pull/3160)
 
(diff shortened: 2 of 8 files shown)


iche033 (approved):

alexdewar:
I might have missed something here... Does this revert fix something?

Before the boost::bind changes there wasn't actually anything *broken* downstream (e.g. in the AUR package I maintain), just a lot of log spam from harmless deprecation warnings, but it seemed like a sensible thing to fix.

alexdewar:
If this fix no longer works, would it be possible to merge my (now-closed) PR #3147 instead? That would at least get rid of the log spam (which can make it harder to debug other issues).

scpeters:
> If this fix no longer works, would it be possible to merge my (now-closed) PR #3147 instead? That would at least get rid of the log spam (which can make it harder to debug other issues).

does https://github.com/osrf/gazebo/pull/3160/commits/1718514d0218283528f29c9f84ab44846b27937d work for you @alexdewar ?

Issue #3159, opened 2022-01-14
Cannot change the heightmap while running


woensug-choi:
## Objective that can't be achieved
Remove the old heightmap and insert a new heightmap in the scene while running. 

## To reproduce
- From gazebo client
Open any world with heightmap (maybe [this tutorial](http://gazebosim.org/tutorials/?tut=dem))
Nothing in the scene changes even with 'remove model' on the gazebo client panel.

- From the gazebo plugin
The function I am using from the gazebo plugin side are
```
world->InsertModelFile('ModelName')
world->RemoveModel('ModelName')
```
which should be a generic method but nothing changes on the screen.

## Background
This feature is essential, when we are working with maritime vehicles operating in a large area (10~50km range). How the [dave project](https://github.com/Field-Robotics-Lab/dave) implemented is to dissect the large GeoTiff bathymetry into tiles and auto spawn and remove according to the vehicle location. We've been doing that by generating the mesh (obj) based model for each tile but wanted to migrate to using the heightmap rendering/physics (which looks much appealing with the Ogre engine at the backend).

https://github.com/Field-Robotics-Lab/dave/pull/178

@scpeters @WilliamLewww

scpeters:
In my testing, it appears that the heightmap collisions are deleted from physics, but the visuals are not deleted from the rendering system.

it seems that `gazebo::rendering::Heightmap` [does not inherit from `gazebo::rendering::Visual`](https://github.com/osrf/gazebo/blob/gazebo11_11.10.0/gazebo/rendering/Heightmap.hh#L62) and requires [special handling with a dummy Visual when being added to a Scene](https://github.com/osrf/gazebo/blob/gazebo11_11.10.0/gazebo/rendering/Scene.cc#L2843-L2848).

I think the logic for responding to `"entity_delete"` requests in [Scene::ProcessRequestMsg](https://github.com/osrf/gazebo/blob/gazebo11_11.10.0/gazebo/rendering/Scene.cc#L2564-L2583) needs to be updated to explicitly remove the heightmap from the scene. It may be as simple as copying [two lines from `Scene::Clear()](https://github.com/osrf/gazebo/blob/gazebo11_11.10.0/gazebo/rendering/Scene.cc#L301-L302).

scpeters:
[...]

Issue #2604, opened 2019-03-08
Heightmap caching of geometry / textures is not documented, indicated, or overridable


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/2604)) by Tully Foote (Bitbucket: [Tully Foote](https://bitbucket.org/%7B3d90dbd4-b845-466f-b5bd-068566f7ff50%7D/), GitHub: [tfoote](https://github.com/tfoote)).**

----------------------------------------
I was generating textured height maps and was having a really hard time debugging them since it seemed that anything I did didn't fix the rendering issue. Then I learned that the output was being cached. Which is why i would see intermittent changes but only if I changed the object name. 

It would be great to see:
 - that it's documented that caching is the default behavior. It's not mentioned in http://gazebosim.org/tutorials?tut=dem
- It would be good to print something in at least the verbose console output that this is happening
- It would be great to have an option to turn off the caching

As a workaround I've been running 

`rm -rf ~/.gazebo/paging/MODELNAME  && gazebo ...` when testing.

scpeters:
I've been having some issues with cached heightmaps not loading textures properly when a different installation of gazebo is used. For example, if I use `/usr/bin/gazebo` from a debian package to load a heightmap for the first time, it will be cached in `~/.gazebo/paging`. If I then open the same heightmap with gazebo compiled from source in a `colcon` workspace in my home folder, it will fail to load the textures and give errors in the `~/.gazebo/ogre.log` file complaining that textures from `/usr/share/gazebo-11` could not be found, since they aren't in the proper ogre resource path. It looks like the checksum used for the cache only considers the `heights` data structure and nothing about the texture paths

scpeters:
cc @tfoote @iche033

scpeters:
I too have found deleting the `~/.gazebo/paging` folder the only way to fix the cache

traversaro:
> If I then open the same heightmap with gazebo compiled from source in a `colcon` workspace in my home folder, it will fail to load the textures and give errors in the `~/.gazebo/ogre.log` file complaining that textures from `/usr/share/gazebo-11` could not be found, since they aren't in the proper ogre resource path. It looks like the checksum used for the cache only considers the `heights` data structure and nothing about the texture paths

[...]

PR #3171, opened 2022-02-11
Scene: support deletion of Heightmap Visuals

scpeters:
As noted in #3159, removing a model containing a heightmap causes its collision shape to be remove from the physics subsystem in `gzserver`, but the heightmap visual remains in `gzclient`. This pull request fixes that so that height map visuals should now disappear from `gzclient` when a heightmap model is deleted. The sequence of commits is described below:

* https://github.com/osrf/gazebo/commit/1445ab1d995cfbcaa425936768b4dce9e0cd5504: This was my first simple attempt to support removing the visual by storing the visual Id associated with the heightmap / terrain and adding logic in `Scene::RemoveVisual` to delete the terrain object when a visual matching `terrainVisualId` is removed. This didn't actually work.
* https://github.com/osrf/gazebo/commit/e497d4d41ae87ba5c2c613f684b1d14b480a60c7: This includes two fixes needed to allow removal of the heightmap visual: using a proper parent visual and adding the visual to the `dataPtr::visuals` vector. This code was already present elsewhere in `ProcessVisualMsg`, so I duplicated it to apply to the heightmap codepath. I didn't like this code duplication, so I worked to eliminate that in the following two commits.
* https://github.com/osrf/gazebo/commit/f3eaaeee12bdb884eb37fb8567045fae6f679015: This changes the Heightmap visual loading to use `Visual::LoadFromMsg` for consistency with the latter parts of `ProcessVisualMsg`. It requires construction of an extra `boost::shared_ptr` but I think the consistency is worth it.
* https://github.com/osrf/gazebo/commit/77bbc8c0814d6fb6d1d2c359696f6b9e860fe08c: This deduplicates the code by moving much of the heightmap processing code to the end of `ProcessVisualMsg` so it can share the existing portions that it had duplicated. This commit has a net reduction of 25 lines.

### DIFF
--- a/gazebo/rendering/Scene.cc
+++ b/gazebo/rendering/Scene.cc
@@ -300,6 +300,7 @@ void Scene::Clear()
 
   delete this->dataPtr->terrain;
   this->dataPtr->terrain = NULL;
+  this->dataPtr->terrainVisualId.reset();
 
   while (!this->dataPtr->visuals.empty())
     this->RemoveVisual(this->dataPtr->visuals.begin()->first);
@@ -2847,6 +2848,7 @@ bool Scene::ProcessVisualMsg(ConstVisualPtr &_msg, Visual::VisualType _type)
         m.clear_material();
         visual->Load(msgs::VisualToSDF(m));
 
+        this->dataPtr->terrainVisualId.emplace(visual->GetId());
         this->dataPtr->terrain = new Heightmap(shared_from_this());
         // check the material fields and set material if it is specified
         if (_msg->has_material())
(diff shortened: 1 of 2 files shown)


scpeters:
> The heightmap will never be visited because it isn't part of `visuals`.

it should be fixed by https://github.com/osrf/gazebo/commit/e497d4d41ae87ba5c2c613f684b1d14b480a60c7

WilliamLewww (approved):
Test failures irrelevant to changes.

tekfr34k:
Thank you for fixing this as it's a functionality I need for a project. I might've spotted a bug however: 
When using ROS for:
- spawning an SDF model to add a heightmap into the scene
- deleting the heightmap from the scene again 
- creating a new heightmap image 
- spawning the SDF model again with the new heightmap (same name, no changes in config) into the scene

For some reason it uses the old visual representation (of the old heightmap) but the collisions work as intended (they're based on the new heightmap). 
A dirty work around that I'm using now is just to spawn the same new heightmap, delete it, then spawn it again, so that "buffer"-like anomaly disappears.

tekfr34k:
Seems my previously described issue was caused by the cache problems with heightmaps similar to the comments in #2604. Deleting the `~/.gazebo/paging/MODELNAME` folder within my script, each time I spawn a new heightmap fixed the problem without needing the workaround of double spawning (which didn't work that great after further testing).

scpeters:
@tekfr34k can you open an issue describing how to reproduce the problem?

thanks!

tekfr34k:
@scpeters I couldn't narrow down what exactly the problem is yet. The issue is very inconsistent, depending on if I start Gazebo from my launch scripts or just without any configuration at all it behaves differently. Maybe it has something to do with defining multiple model paths (even if the models are named differently in the SDF/config). Adding a model path in the "Insert" tab which contains the heightmap SDF gives an issue of not showing the model in the overview. Maybe that's something to start with...

PR #3187, opened 2022-03-06
Support plotting for entities with / in the name

### LINKED ISSUE #3181: Plotting should show complete joint names that contain slashes
I have been debugging an issue with the [dave_manipulator_spring_demo](https://github.com/Field-Robotics-Lab/dave/blob/master/examples/dave_demo_launch/launch/dave_manipulator_spring_demo.launch) and wanted to plot some joint velocities but noticed that the joint names were not showing up correctly. Currently, the joint names in the robot model from this launch file uses the following pattern: `namespace/unique_joint_name`, but only the portion of the name before the `/` shows up in the plot window.

I think we could use URL encoding to handle reserved characters to resolve this. For example, this would replace any `/` characters in joint names with `%2f`. One option is to use [curl_easy_escape](https://curl.se/libcurl/c/curl_easy_escape.html) and [curl_easy_unescape](https://curl.se/libcurl/c/curl_easy_escape.html) to encode and decode the relevant names. Note that these APIs allocate new memory that must be cleared with `curl_free`.

scpeters:
I noticed while testing the rexrov models from
Field-Robotics-Lab/dave that the plotting window
does not work properly when elements of a model include
a forward-slash `/` in the entity name. To fix the problem,
the `/` characters are encoded as %2f in physics::Base::URI().
Corresponding decoding is added to the gui/plot/Palette class.

Fixes #3181.

### DIFF
--- a/gazebo/physics/Base.cc
+++ b/gazebo/physics/Base.cc
@@ -14,6 +14,7 @@
  * limitations under the License.
  *
 */
+#include <boost/algorithm/string.hpp>
 #include "gazebo/common/Assert.hh"
 #include "gazebo/common/Console.hh"
 #include "gazebo/common/Exception.hh"
@@ -383,7 +384,9 @@ common::URI Base::URI() const
   {
     if (p->GetParent())
     {
-      uri.Path().PushFront(p->GetName());
+      std::string escapedParentName = p->GetName();
+      boost::replace_all(escapedParentName, "/", "%2f");
+      uri.Path().PushFront(escapedParentName);
       uri.Path().PushFront(p->TypeStr());
     }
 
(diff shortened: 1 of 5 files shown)


gazebo/physics/Base.cc
```diff
@@ -383,15 +384,19 @@ common::URI Base::URI() const
   {
     if (p->GetParent())
     {
-      uri.Path().PushFront(p->GetName());
+      std::string escapedParentName = p->GetName();
+      boost::replace_all(escapedParentName, "/", "%2f");
```
iche033: use `common::replaceAll` instead of `boost`? I do still see `boost` being used in gazebo so maybe not a big deal

iche033 (approved):
looks good, test passes for me. Just one minor comment but PR is fine as is.

gazebo/physics/Base.cc
scpeters: I switched to `common::replaceAll` in https://github.com/osrf/gazebo/pull/3187/commits/7edb013c5c86738ea6489609f042b099f9707c67

PR #3200, opened 2022-03-24
Separate cache files for each heightmap LOD

### LINKED ISSUE #3199: Heightmap paging cache corruption when mixing LOD values
## Background

The `gazebo::rendering::Heightmap` class currently caches height values and writes them to the `~/.gazebo/paging` folder and reads from this cache to speed up future loads. Some problems have been reported with this cache (such as https://github.com/osrf/gazebo/issues/2604), and a common workaround is to delete the entire `~/.gazebo/paging` folder.

## Summary of current bug

We have noticed that the Heightmap rendering can be distorted when loading large heightmaps (4096x4096 and larger) with both mixed zero and nonzero values of LOD in `gzserver` and `gzclient`. I believe this is related to the changes from [bitbucket PR 2655](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/2655/page/1) (merged in https://github.com/osrf/gazebo/commit/b5f39ee5949396951c260f993b4508a435485389). Some observations:

* When using `gzserver` to load a large heightmap with `LOD == 0` with a freshly cleared `~/.gazebo/paging` folder, it generates a new subfolder containing a `gzterrain.SHA1` and a single `gazebo_terrain_00000000.dat`.
  - When loading `gzclient` with a matching value of `LOD == 0`, rendering is fine.
  - When loading `gzclient` with a nonzero value of LOD, rendering is distorted.
* When using a `gzserver` to load a large heightmap with `LOD > 0` with a freshly cleaned `~/.gazebo/paging` folder, it generates a new subfolder containing a `gzterrain.SHA1` and multiple `gazebo_terrain_*.dat` files.
  - When loading `gzclient` also with a nonzero value of LOD, rendering is fine.
  - When loading `gzclient` with `LOD == 0`, rendering is distorted.

So perhaps we should be creating separate cache folders for `LOD == 0`  and `LOD > 0`? Or should we not use the cache if `LOD == 0`?

WilliamLewww:
Closes #3199 

To prevent gazebo from loading the wrong cache files, each heightmap LOD will have its own set of cache files.

The LOD is appended to the cache files' names (_5_ in the example):
```
[Msg] Loading heightmap cache data: /root/.gazebo/paging/heightmap_bowl_2/gazebo_terrain_5_00000000.dat
[Msg] Loading heightmap cache data: /root/.gazebo/paging/heightmap_bowl_2/gazebo_terrain_5_00010000.dat
[Msg] Loading heightmap cache data: /root/.gazebo/paging/heightmap_bowl_2/gazebo_terrain_5_00000001.dat
[Msg] Loading heightmap cache data: /root/.gazebo/paging/heightmap_bowl_2/gazebo_terrain_5_00010001.dat
```

### DIFF
--- a/gazebo/rendering/Heightmap.cc
+++ b/gazebo/rendering/Heightmap.cc
@@ -550,7 +550,7 @@ void Heightmap::Load()
       gzmsg << "Large heightmap used with LOD. It will be subdivided into " <<
           this->dataPtr->numTerrainSubdivisions << " terrains." << std::endl;
     }
-    std::string terrainName = "gazebo_terrain_" + terrainNameSuffix;
+    std::string terrainName = "gazebo_terrain" + terrainNameSuffix;
     prefix = terrainDirPath / terrainName.c_str();
   }
 
--- a/gazebo/rendering/Heightmap.cc
+++ b/gazebo/rendering/Heightmap.cc
@@ -518,13 +518,19 @@ void Heightmap::Load()
     }
   }
 
+  std::string terrainNameSuffix = "";
+  if (this->LOD() == 0)
+  {
+    terrainNameSuffix = "_LOD0";
+  }
+
   // If the paging is enabled we modify the number of subterrains
   if (this->dataPtr->useTerrainPaging)
   {
     this->dataPtr->splitTerrain = true;
     nTerrains = this->dataPtr->numTerrainSubdivisions;
     std::string terrainName = "gazebo_terrain_cache" +
-        std::to_string(this->LOD());
+        terrainNameSuffix;
     prefix = terrainDirPath / terrainName.c_str();
   }
   else
@@ -544,7 +550,7 @@ void Heightmap::Load()
       gzmsg << "Large heightmap used with LOD. It will be subdivided into " <<
           this->dataPtr->numTerrainSubdivisions << " terrains." << std::endl;
     }
-    std::string terrainName = "gazebo_terrain_" + std::to_string(this->LOD());
+    std::string terrainName = "gazebo_terrain_" + terrainNameSuffix;
     prefix = terrainDirPath / terrainName.c_str();
   }
 
--- a/gazebo/rendering/Heightmap.cc
+++ b/gazebo/rendering/Heightmap.cc
@@ -523,7 +523,9 @@ void Heightmap::Load()
   {
     this->dataPtr->splitTerrain = true;
     nTerrains = this->dataPtr->numTerrainSubdivisions;
-    prefix = terrainDirPath / "gazebo_terrain_cache";
+    std::string terrainName = "gazebo_terrain_cache" +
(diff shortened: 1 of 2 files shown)


scpeters:
> Closes #3199
> 
> To prevent gazebo from loading the wrong cache files, each heightmap LOD will have its own set of cache files.
> 
> The LOD is appended to the cache files' names (_5_ in the example):
> 
> ```
> [Msg] Loading heightmap cache data: /root/.gazebo/paging/heightmap_bowl_2/gazebo_terrain_5_00000000.dat
> [Msg] Loading heightmap cache data: /root/.gazebo/paging/heightmap_bowl_2/gazebo_terrain_5_00010000.dat
> [Msg] Loading heightmap cache data: /root/.gazebo/paging/heightmap_bowl_2/gazebo_terrain_5_00000001.dat
> [Msg] Loading heightmap cache data: /root/.gazebo/paging/heightmap_bowl_2/gazebo_terrain_5_00010001.dat
> ```

thanks I'll go ahead and test this. It's possible that cache files are equivalent for any nonzero LOD values, in which case we could simplify this a bit, but I expect that this will fix the issue

gazebo/rendering/Heightmap.cc
```diff
@@ -523,7 +523,9 @@ void Heightmap::Load()
   {
     this->dataPtr->splitTerrain = true;
     nTerrains = this->dataPtr->numTerrainSubdivisions;
-    prefix = terrainDirPath / "gazebo_terrain_cache";
+    std::string terrainName = "gazebo_terrain_cache" +
+        std::to_string(this->LOD());
```
scpeters: I've noticed that the cache files appear to be identical for any `LOD > 0`. So I don't think we need to use separate file names for each `LOD` value, just one for `LOD == 0` and one for `LOD > 0`.

How about the following?

~~~
std::string terrainNameSuffix;
if (this->LOD() == 0)
{
  terrainNameSuffix = "_LOD0";
}
std::string terrainName = "gazebo_terrain_cache" + terrainNameSuffix;
~~~

also, to avoid code duplication `terrainNameSuffix` could also be computed before this `if` branch so it could also be used in the `else` branch

scpeters (commented):
> thanks I'll go ahead and test this. It's possible that cache files are equivalent for any nonzero LOD values, in which case we could simplify this a bit, but I expect that this will fix the issue

I've confirmed that this fixes the issue, but also that the cache files appear to be identical for any `LOD > 0`. I've suggested a code change inline that may reduce the number of cache files

mogumbo:
Tested some big terrains using gzserver lod = 0 and gzclient lod = 3. Terrain looked good in both gzclient and cameras. Without this PR, the structure of the terrain had visible flaws.

scpeters (approved):

PR #3219, opened 2022-05-18
Get/Set friction coefficients of wheels

jennuine:
Extends wheel slip plugin to get and set individual wheel friction coefficients in the primary and secondary directions

### DIFF
--- a/plugins/WheelSlipPlugin.cc
+++ b/plugins/WheelSlipPlugin.cc
@@ -482,6 +482,54 @@ void WheelSlipPlugin::SetSlipComplianceLongitudinal(std::string _wheel_name, con
 }
 
 /////////////////////////////////////////////////
+std::map<std::string, ignition::math::Vector2d> WheelSlipPlugin::GetFrictionCoefficients()
+{
+  std::lock_guard<std::mutex> lock(this->dataPtr->mutex);
+  std::map<std::string, ignition::math::Vector2d> frictionCoeffs;
+  for (const auto &linkSurface : this->dataPtr->mapLinkSurfaceParams)
+  {
+    auto link = linkSurface.first.lock();
+    auto surface = linkSurface.second.surface.lock();
+    if (!link || !surface)
+      continue;
+
+    ignition::math::Vector2d friction;
+    friction.X(surface->FrictionPyramid()->MuPrimary());
+    friction.Y(surface->FrictionPyramid()->MuSecondary());
+
+    frictionCoeffs[link->GetName()] = friction;
+  }
+
+  return frictionCoeffs;
+}
+
+/////////////////////////////////////////////////
+void WheelSlipPlugin::SetMuPrimary(const std::string &_wheel_name, const double _mu)
+{
+  std::lock_guard<std::mutex> lock(this->dataPtr->mutex);
+  if (this->dataPtr->mapLinkNames.count(_wheel_name) > 0)
+  {
+    auto link = this->dataPtr->mapLinkNames[_wheel_name];
+    auto surface = this->dataPtr->mapLinkSurfaceParams[link].surface.lock();
+    if (surface)
+      surface->FrictionPyramid()->SetMuPrimary(_mu);
+  }
+}
+
+/////////////////////////////////////////////////
+void WheelSlipPlugin::SetMuSecondary(const std::string &_wheel_name, const double _mu)
+{
+  std::lock_guard<std::mutex> lock(this->dataPtr->mutex);
+  if (this->dataPtr->mapLinkNames.count(_wheel_name) > 0)
+  {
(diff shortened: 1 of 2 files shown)


plugins/WheelSlipPlugin.hh
```diff
@@ -151,6 +151,22 @@ namespace gazebo
...
+    /// brief Get friction coefficients for each wheel
+    /// \return Map of wheel name to a Vector2 of friction coefficients
+    /// The Vector2.X value is the friction coefficient in the primary direction and
+    /// the Vector2.Y value is the friction coefficient in the secondary direction.
+    public: std::map<std::string, ignition::math::Vector2d> GetFrictionCoefficients();
+
+    /// \brief Set the friction coefficient in the primary direction for a particular wheel.
+    /// \param[in] _wheel_name name of the wheel link on which _mu should be set.
+    /// \param[in] _mu Friction coefficient.
+    public: void SetMuPrimary(const std::string &_wheel_name, const double _mu);
```
jacobperron: nitpick: making `_mu` const doesn't mean much to me since we're passing by value. I would remove the `const`.

plugins/WheelSlipPlugin.cc
```diff
@@ -481,6 +481,54 @@ void WheelSlipPlugin::SetSlipComplianceLongitudinal(std::string _wheel_name, con
...
+  }
+
+  return frictionCoeffs;
+}
+
+/////////////////////////////////////////////////
+void WheelSlipPlugin::SetMuPrimary(const std::string &_wheel_name, const double _mu)
+{
+  std::lock_guard<std::mutex> lock(this->dataPtr->mutex);
+  if (this->dataPtr->mapLinkNames.count(_wheel_name) > 0)
```
jacobperron: I'm not sure what the convention is in Gazebo for handling errors, but maybe we should at least log an error in the case that `wheel_name` is invalid?

jacobperron (approved):
LGTM, I have a few minor comments

plugins/WheelSlipPlugin.hh
scpeters: agree and same for `SetMuSecondary`

plugins/WheelSlipPlugin.cc
scpeters: this is copying the approach in `SetSlipComplianceLongitudinal` from earlier in this file. If we add a console message, we should be careful that it doesn't print for every simulation time step and flood the console

jacobperron: I don't think it's necessarily wrong to quietly ignore invalid wheel names, I was just wondering if there was a preference. Since there is precedent with the slip compliance API, we could leave this as-is.

scpeters: we could alternatively return `bool` from the new `Set*` functions and return false if the name is not recognized

jennuine: > we could alternatively return bool from the new Set* functions and return false if the name is not recognized

Would it be odd to not match `SetSlipCompliance*`?

scpeters: a little odd, but if we prefer that approach, then we should do it

jennuine: > I'm not sure what the convention is in Gazebo for handling errors, but maybe we should at least log an error in the case that wheel_name is invalid?

> this is copying the approach in SetSlipComplianceLongitudinal from earlier in this file. If we add a console message, we should be careful that it doesn't print for every simulation time step and flood the console

I'm leaning towards leaving it as a `void` function, logging an error when the `wheel_name` is not found, and returning. I don't think we need to worry about flooding the console because user's shouldn't be updating the parameters at every sim step. Thoughts?

jacobperron: Perhaps not in our use-case, but I can imagine a scenario where a plugin calls this API at every simulation step. I think what Steve suggests makes sense because then the user can decide whether or not to log an error. 

I don't know if it's too late to update the other setters (e.g. we probably don't want to modify the API if it's already been released); but it's something to keep in mind when porting this to Gazebo (next generation).

scpeters: I prefer returning `bool` for the new API, even thought it deviates from the existing API pattern

plugins/WheelSlipPlugin.hh
jennuine: Removed 466faaa80159e5b473bcd745078ffc83e172fddc

plugins/WheelSlipPlugin.cc
jennuine: Done 466faaa80159e5b473bcd745078ffc83e172fddc

scpeters (approved):

PR #3226, opened 2022-06-07
Updated README links

jennuine:
Updated README links to point to gazebo-classic pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,13 @@ Gazebo - A dynamic multi-robot simulator
 This is the Gazebo simulator.  Gazebo simulates multiple robots in a
 3D environment, with extensive dynamic interaction between objects.
 
-  http://gazebosim.org
+  http://classic.gazebosim.org
 
 Installation
 ------------
 Instructions are located at
 
-  http://gazebosim.org/install
+  http://classic.gazebosim.org/install
 
 Gazebo cmake parameters available at configuring time:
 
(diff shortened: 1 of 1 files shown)


quarkytale:
Install link is still broken, this [install from source link](https://classic.gazebosim.org/tutorials?tut=install_from_source&cat=install#InstallGazebofromsource(UbuntuandMac)) works, can we add it somehow?

README.md
```diff
@@ -4,13 +4,13 @@ Gazebo - A dynamic multi-robot simulator
...
 
-  http://gazebosim.org
+  http://classic.gazebosim.org
 
 Installation
 ------------
 Instructions are located at
 
-  http://gazebosim.org/install
+  http://classic.gazebosim.org/install
```
iche033, line 13: I think this link is broken? Could be an issue with the website

jennuine: Yes, unfortunately it has been broken for awhile. There is a slack thread regarding this. How about I leave the link there for when it gets fixed and add the install from source link @quarkytale suggested?

quarkytale: I'm curious how the uninstall link goes to 'uninstall from source' but not the install one

quarkytale (commented):
LGTM apart from the install link, [slack thread for reference ](https://osrf.slack.com/archives/C8LMU4733/p1654281183352419)

I'm curious as why the uninstall link directs to 'uninstall from source' but not the install one

README.md
iche033: here's the redirect: https://github.com/osrf/gazebo_web_app/blob/master/config/routes.rb#L48

iche033: we probably just need to update the redirect link there

iche033 (approved):
looks good to me. The broken gazebo install link is a separate issue.

README.md
quarkytale: I can take that up

jennuine:
Here's the issue for cross-referencing: https://github.com/osrf/gazebo_web_app/issues/38

PR #3227, opened 2022-06-09
Add missing header files

scpeters:
Some header files are proposed for removal from `sdf/Param.hh` in https://github.com/gazebosim/sdformat/pull/1044. While testing that branch with gazebo11, I noticed some compilation failures in files that use objects without including the associated header file. I've added missing header files for many `ignition/math/*` includes in https://github.com/osrf/gazebo/commit/c20bb21f4c282df7207da74b668911d719bd1a0c. It is more than enough to fix compilation, though it is not fully exhaustive.

I made two other small changes as well:

* https://github.com/osrf/gazebo/commit/d51a7de774c303db427f93ebcc8e23ccf1f9db47: fix a spelling error
* https://github.com/osrf/gazebo/commit/d5f9d42ca6e8b9078d58959bb3f3441760528d6e: replace inclusion of `ignition/math.hh` in two `.cc` files with the specific math header files that are needed to reduce compilation time. Since these are `.cc` files, there should not be an effect on any downstream code.

### DIFF
--- a/examples/stand_alone/marker/marker.cc
+++ b/examples/stand_alone/marker/marker.cc
@@ -15,7 +15,9 @@
  *
 */
 #include <ignition/transport.hh>
-#include <ignition/math.hh>
+#include <ignition/math/Pose3.hh>
+#include <ignition/math/Rand.hh>
+#include <ignition/math/Vector3.hh>
 #include <ignition/msgs.hh>
 #include <gazebo/common/Time.hh>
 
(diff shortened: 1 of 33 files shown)


scpeters:
https://github.com/gazebosim/sdformat/pull/1044 has been closed, but these changes are still valid, so I will leave this open for review

j-rivero:
> [gazebosim/sdformat#1044](https://github.com/gazebosim/sdformat/pull/1044) has been closed, but these changes are still valid, so I will leave this open for review

Missed the original discussion but Louise was right about not removing a header from an stable released series of sdformat.

The PR looks good to me, CI is happy, I'm happy too.

j-rivero (approved):

j-rivero:
There is a problem on Windows conda-forge but I'm pretty sure it is not related to this PR.

PR #3164, opened 2022-01-22
Wheel plowing approximation in ODEPhysics

scpeters:
As noted in https://github.com/osrf/gazebo/issues/3085, the distribution of contact forces can be quite different for a wheel driving on rigid terrain and deformable terrain.  Contact for rigid wheel on rigid terrain can be modeled with a single contact point and normal direction perpendicular to the surface. For a rigid wheel on deformable terrain, the contact occurs over a larger area. The contact pressure distribution can be approximated as an equivalent point force at the centroid of the pressure distribution. At rest, a wheel on a deformable surface may sink into the surface with a symmetric distribution of normal contact pressure, such that the equivalent normal force vector is parallel to the undisturbed surface normal, which is vertical in the illustration below.

wheel_plowing

While rolling, the sinkage of the wheel may cause a plowing effect, in which the terrain material is bunched up in the direction of travel and compacted in its wake. This can cause a significant forward bias in the distribution of normal contact pressure and an equivalent normal force vector that is inclined relative to the undisturbed surface normal. In the illustration above, the equivalent normal force opposes the motion of the wheel, requiring tractive torque to be applied in order to maintain a steady velocity. Even when flat, the plowing effect means that driving on a deformable surface is energetically equivalent to driving uphill on a rigid surface.

This pull request implements an approximation of longitudinal wheel plowing effects by displacing contact points and inclining the contact normals to replicate the behavior illustrated in the figure above. This is implemented in `ODEPhysics::Collide`, which is a callback function called for each pair of colliding bodies. To enable the plowing effect, special SDFormat tags must be added to the terrain (`<gz:plowing_terrain />`) and the wheels (`<gz:plowing_wheel />`). The wheel must specify `<fdir1>` in its `<surface>`, and the `<gz:plowing_wheel>` tag must also contain `<max_degrees>` and `<saturation_velocity>` and may optionally contain `<deadband_velocity>`. See the `plowing_effect_*` models in `test/models` for examples.

PR #3229, opened 2022-06-15
plowing effect unit test

deepanshubansal01:
This PR implements unit test for plowing effect feature, more details here  #3164

### DIFF
--- a/test/integration/plowing_effect.cc
+++ b/test/integration/plowing_effect.cc
@@ -199,6 +199,8 @@ void PlowingEffectSpheres::CallbackMaxPlowingAngle(const ConstContactsPtr &_msg)
 
 void PlowingEffectTricycle::RigidTerrain(const std::string &_physicsEngine)
 {
+  gazebo::common::SystemPaths::Instance()->AddModelPaths(
+                    PROJECT_SOURCE_PATH "/test/models");
   // Load the plowing effect world
   Load("worlds/plowing_effect_demo.world", true, _physicsEngine);
   physics::WorldPtr world = physics::get_world("default");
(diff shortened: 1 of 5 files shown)


test/integration/plowing_effect.cc
```diff
@@ -0,0 +1,197 @@
...
+      {
+        EXPECT_EQ(angle, 0);
+      }
+
+      // plowing effect
+      else if(wheelSpeedLongitudinal > deadbandVelocity_ && wheelSpeedLongitudinal < saturationVelocity_)
+      {
+          // angle less than maxRadians_
+          ASSERT_LT(angle, maxRadians_);
+        // TODO: should we also check for exact angle values here
```
deepanshubansal01: @adityapande-1995 for this should I compute the exact angle using the same logic used in ODEPhysics?

adityapande-1995: We could calculate the angle and use EXPECT_NEAR macro to test, by including some tolerance.

test/worlds/plowing_effect_test.world
```diff
@@ -0,0 +1,196 @@
...
+
+    <model name="sphere1">
+      <link name='base_link'>
+        <pose>0 0 0 0 0 0</pose>
+        <inertial>
+          <mass>0.5</mass>
+          <inertia>
+            <ixx>0.22799999999999998</ixx>
+            <iyy>0.7435210984814149</iyy>
+            <izz>0.9655210984814149</izz>
```
scpeters: I would expect the `ixx`, `iyy` and `izz` values to be equal for a sphere

scpeters (commented):
thanks, I will try testing this. In the meantime I noticed something odd in the sphere inertias in the test world

test/worlds/plowing_effect_test.world
deepanshubansal01: Thanks, it's corrected now.

deepanshubansal01:
@scpeters @adityapande-1995 I have fixed the issue with model paths. The unit tests should run now without exporting the `GAZEBO_MODEL_PATH` env variable.

scpeters:
thanks, the tests are looking great! I'm going to make a few changes, but first I'll merge this into #3164

scpeters (approved):

deepanshubansal01:
> thanks, the tests are looking great! I'm going to make a few changes, but first I'll merge this into #3164

Awesome, thanks for the review:)

PR #3235, opened 2022-06-30
Support shininess value for each Visual in a Model

### LINKED ISSUE #3232: Feature Request: support different shininess values for multiple visuals in a model
## Background

A feature for applying a shininess value to visuals from an SDFormat file was added in https://github.com/osrf/gazebo/pull/3213. In Gazebo11, the SDFormat world file is parsed by `physics::World` and communicated to the `rendering::Scene` over a transport interface. While it would have been simplest to add a `shininess` field to the relevant protobuf message, this would have caused ABI problems. As a workaround, an ignition service is created for each model that can be called by the Scene to retrieve the shininess of the first link of the first visual of that model (see this portion of [World::LoadModel](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/physics/World.cc#L1264-L1283)). Although shininess can be specified on a per-visual basis in SDFormat, it is limited by the transport service mechanism to only report one shininess value per model.

## Feature request

Expand the functionality of the service interface so that the shininess of each visual in a model can be queried independently.

scpeters:
As noted in #3232, only a single shininess value is currently supported for each model. This pull request expands the functionality of the ignition service interface so that the shininess of each Visual in a Model can be queried independently. This works by storing the shininess values by the scoped visual name. It also replaces the per-model ignition services with a single `/shininess` service.

It is technically a change to the transport interface, but since this is such a new feature and improved by the current changes, I think it is an acceptable behavior change.

Fixes #3232.

### DIFF
--- a/gazebo/physics/Link.cc
+++ b/gazebo/physics/Link.cc
@@ -1547,6 +1547,16 @@ void Link::UpdateVisualMsg()
       bool newVis = true;
       std::string visName = this->GetScopedName() + "::" + msg.name();
 
+      if (visualElem->HasElement("material"))
+      {
+        sdf::ElementPtr matElem = visualElem->GetElement("material");
+        if (matElem->HasElement("shininess"))
+        {
+          this->world->SetVisualShininess(
+              visName, matElem->Get<double>("shininess"));
+        }
+      }
+
       // update visual msg if it exists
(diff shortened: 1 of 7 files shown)


jacobperron (approved):
LGTM

PR #3245, opened 2022-07-27
Camera: expose intrinsics parameters

deepanshubansal01:
This PR makes efforts to expose camera intrinsic parameters so that they can be accessed by `gazebo_ros_camera` plugin. 
Currently, the Camera API exposes a `ProjectionMatrix()` which is normalised based on Open GL graphics and whose  values doesn't make sense to the user using the API. 

The PR exposed `ImageFocalLengthX(), ImageFocalLengthY(), ImageOpticalCentreX() and ImageOpticalCentreY()` API's. These values are populated from the `intrinsic` tag in case intrinsic are provided in `sdf` file. If no intrinsic are provided then these values are decoupled from the `ProjectionMatrix` since the ProjectionMatrix is calculated by default based on `fov`, `image size` and we can decouple it to get the intrinsic.

Some related discussion here: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1407

Signed-off-by: deepanshu <deepanshubansal01@gmail.com>

### DIFF
--- a/gazebo/rendering/Camera.cc
+++ b/gazebo/rendering/Camera.cc
@@ -121,8 +121,7 @@ Camera::Camera(const std::string &_name, ScenePtr _scene,
 
   this->dataPtr->antiAliasingValue = 4;
 
-  // calculate intrinsics from default matrix
-  this->CalculateIntrinsicsFromProjectionMatrix();
+  this->dataPtr->cameraIntrinsicMatrix = ignition::math::Matrix3d::Identity;
 }
 
 //////////////////////////////////////////////////
@@ -238,8 +237,8 @@ void Camera::UpdateCameraIntrinsics(
     _cameraIntrinsicsCx, _cameraIntrinsicsCy,
     _cameraIntrinsicsS, clipNear, clipFar);
   this->dataPtr->cameraIntrinsicMatrix = this->BuildIntrinsicMatrix(
-              _cameraIntrinsicsFx, _cameraIntrinsicsFy,
-              _cameraIntrinsicsCx, _cameraIntrinsicsCy);
+             _cameraIntrinsicsFx, _cameraIntrinsicsFy,
+             _cameraIntrinsicsCx, _cameraIntrinsicsCy);
 
   if (this->camera != nullptr)
   {
@@ -2196,8 +2195,8 @@ void Camera::CalculateIntrinsicsFromProjectionMatrix()
   double inverseWidth = 1.0 / (right - left);
   double inverseHeight = 1.0 / (top - bottom);
 
-  double fX = m(0, 0)/ 2*inverseWidth;
-  double fY = m(1, 1) / 2*inverseHeight;
+  double fX = m(0, 0)/ (2*inverseWidth);
+  double fY = m(1, 1) / (2*inverseHeight);
   double cX = -((m(0, 2) - ((right + left) / (right - left))) * (right - left)) / 2;
   double cY = this->imageHeight + ((m(1, 2) - ((top + bottom) / (top - bottom))) * (top - bottom)) / 2;
 
--- a/gazebo/rendering/Camera.cc
+++ b/gazebo/rendering/Camera.cc
@@ -237,7 +237,7 @@ void Camera::UpdateCameraIntrinsics(
     _cameraIntrinsicsFx, _cameraIntrinsicsFy,
     _cameraIntrinsicsCx, _cameraIntrinsicsCy,
     _cameraIntrinsicsS, clipNear, clipFar);
-  this->cameraIntrinsicMatrix = this->BuildIntrinsicMatrix(
+  this->dataPtr->cameraIntrinsicMatrix = this->BuildIntrinsicMatrix(
               _cameraIntrinsicsFx, _cameraIntrinsicsFy,
               _cameraIntrinsicsCx, _cameraIntrinsicsCy);
 
@@ -1120,25 +1120,25 @@ double Camera::FarClip() const
 //////////////////////////////////////////////////
 double Camera::FocalLengthX() const
 {
-  return this->cameraIntrinsicMatrix(0, 0);
+  return this->dataPtr->cameraIntrinsicMatrix(0, 0);
 }
 
 //////////////////////////////////////////////////
 double Camera::FocalLengthY() const
 {
-  return this->cameraIntrinsicMatrix(1, 1);
+  return this->dataPtr->cameraIntrinsicMatrix(1, 1);
 }
 
 //////////////////////////////////////////////////
 double Camera::OpticalCentreX() const
 {
-  return this->cameraIntrinsicMatrix(0, 2);
+  return this->dataPtr->cameraIntrinsicMatrix(0, 2);
 }
 
 //////////////////////////////////////////////////
 double Camera::OpticalCentreY() const
 {
-  return this->cameraIntrinsicMatrix(1, 3);
+  return this->dataPtr->cameraIntrinsicMatrix(1, 3);
 }
 
 //////////////////////////////////////////////////
@@ -2201,7 +2201,7 @@ void Camera::CalculateIntrinsicsFromProjectionMatrix()
   double cX = -((m(0, 2) - ((right + left) / (right - left))) * (right - left)) / 2;
   double cY = this->imageHeight + ((m(1, 2) - ((top + bottom) / (top - bottom))) * (top - bottom)) / 2;
 
-  this->cameraIntrinsicMatrix = this->BuildIntrinsicMatrix(
+  this->dataPtr->cameraIntrinsicMatrix = this->BuildIntrinsicMatrix(
       fX, fY, cX, cY);
 }
 
--- a/gazebo/rendering/Camera.cc
+++ b/gazebo/rendering/Camera.cc
@@ -2188,20 +2188,13 @@ void Camera::CalculateIntrinsicsFromProjectionMatrix()
 {
   const ignition::math::Matrix4d& projectionMat = this->ProjectionMatrix();
 
-  double right = this->imageWidth;
-  double left = 0.0;
-  double top = this->imageHeight;
-  double bottom = 0.0;
+  const double width = this->imageWidth;
+  const double height = this->imageHeight;
 
-  double inverseWidth = 1.0 / (right - left);
-  double inverseHeight = 1.0 / (top - bottom);
-
-  double fX = projectionMat(0, 0)/ (2*inverseWidth);
-  double fY = projectionMat(1, 1) / (2*inverseHeight);
-  double cX = -((projectionMat(0, 2) -
-            ((right + left) / (right - left))) * (right - left)) / 2;
-  double cY = this->imageHeight + ((projectionMat(1, 2) -
-            ((top + bottom) / (top - bottom))) * (top - bottom)) / 2;
+  double fX = (projectionMat(0, 0) * width) / 2;
+  double fY = (projectionMat(1, 1) * height) / 2;
+  double cX = -(width * (projectionMat(0, 2) - 1.0)) / 2;
+  double cY = height + (height * (projectionMat(1, 2) - 1.0)) / 2;
 
   this->dataPtr->cameraIntrinsicMatrix = this->BuildIntrinsicMatrix(
       fX, fY, cX, cY);
--- a/gazebo/rendering/Camera.cc
+++ b/gazebo/rendering/Camera.cc
@@ -1117,25 +1117,25 @@ double Camera::FarClip() const
 }
 
 //////////////////////////////////////////////////
-double Camera::FocalLengthX() const
+double Camera::ImageFocalLengthX() const
 {
   return this->dataPtr->cameraIntrinsicMatrix(0, 0);
 }
 
 //////////////////////////////////////////////////
-double Camera::FocalLengthY() const
+double Camera::ImageFocalLengthY() const
 {
   return this->dataPtr->cameraIntrinsicMatrix(1, 1);
 }
 
 //////////////////////////////////////////////////
-double Camera::OpticalCentreX() const
+double Camera::ImageOpticalCentreX() const
 {
   return this->dataPtr->cameraIntrinsicMatrix(0, 2);
 }
 
 //////////////////////////////////////////////////
-double Camera::OpticalCentreY() const
+double Camera::ImageOpticalCentreY() const
 {
   return this->dataPtr->cameraIntrinsicMatrix(1, 3);
 }
--- a/gazebo/rendering/Camera.cc
+++ b/gazebo/rendering/Camera.cc
@@ -245,6 +245,7 @@ void Camera::UpdateCameraIntrinsics(
     this->camera->setCustomProjectionMatrix(true,
         Conversions::Convert(this->cameraProjectiveMatrix));
   }
+
   this->cameraUsingIntrinsics = true;
 }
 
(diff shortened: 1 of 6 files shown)


gazebo/rendering/Camera.hh
```diff
@@ -911,6 +941,9 @@ namespace gazebo
       /// \brief Flag for signaling the usage of camera intrinsics within OGRE
       protected: bool cameraUsingIntrinsics;
 
+      /// \brief Camera Intrinsic Matrix
+      protected: ignition::math::Matrix3d cameraIntrinsicMatrix;
```
scpeters: FYI this breaks ABI. If it can be private instead of `protected`, you can put it in https://github.com/osrf/gazebo/blob/gazebo11/gazebo/rendering/CameraPrivate.hh

adityapande-1995 (commented):
Can we have this test case : We set the intrinsics as the default values, get the interinsics using ``FocalLengthX/Y()``. Now don't set the intrinsics in sdf (for a separate camera), let the sensor use default values, and query the intrinsics again using the new methods. These values should be same as those obtained earlier, and the images should also be the same visually.

deepanshubansal01:
> Can we have this test case : We set the intrinsics as the default values, get the interinsics using `FocalLengthX/Y()`. Now don't set the intrinsics in sdf (for a separate camera), let the sensor use default values, and query the intrinsics again using the new methods. These values should be same as those obtained earlier, and the images should also be the same visually.

Yeah that’s a good test case to have. This test case should be in `gazebo` or `gazebo_ros_pkgs`?

deepanshubansal01:
Should we expose the API's for `ImageFocalLengthX()`, `ImageOpticalCentreX()` etc in `CameraSensor` similar to `ImageWidth()` and `ImageHeight()` API's? cc @adityapande-1995 @jacobperron https://github.com/osrf/gazebo/blob/127108a4137c70e59704281399465efe0e90fc19/gazebo/sensors/CameraSensor.hh#L97-L101

We can also get those values though from `rendering::Camera` object of `CameraSensor` https://github.com/osrf/gazebo/blob/127108a4137c70e59704281399465efe0e90fc19/gazebo/sensors/CameraSensor.hh#L135

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,121 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+      rendering::RenderEngine::NONE)
+  {
+    gzerr << "No rendering engine, unable to run camera test\n";
+    return;
+  }
+
+  // get the 'default_intrinsics_camera' sensor, no <intrinsics> tag is used for this sensor
+  // and we expect the intrinsics to be equal to the default intrinsic values provided by gazebo.
+  std::string defaultIntrinsicsCameraName = "default_intrinsics_camera_sensor";
+  sensors::SensorPtr defaultIntrinsicsSensor = sensors::get_sensor(defaultIntrinsicsCameraName);
```
jacobperron:
```suggestion
  sensors::SensorPtr defaultIntrinsicsSensor = sensors::get_sensor("default_intrinsics_camera_sensor");
```

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,121 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+  sensors::CameraSensorPtr defaultIntrinsicsCamSensor =
+      std::dynamic_pointer_cast<sensors::CameraSensor>(defaultIntrinsicsSensor);
+  EXPECT_TRUE(defaultIntrinsicsCamSensor != nullptr);
+  rendering::CameraPtr defaultIntrinsicsCam = defaultIntrinsicsCamSensor->Camera();
+  EXPECT_TRUE(defaultIntrinsicsCam != nullptr);
+
+  // image size, focal length and optical centre for 'default_intrinsics_camera' sensor
+  unsigned int defaultCamWidth = defaultIntrinsicsCam->ImageWidth();
+  unsigned int defaultCamHeight = defaultIntrinsicsCam->ImageHeight();
+  {
```
jacobperron: nitpick: it doesn't seems like this extra scope is accomplishing anything. I think the code behaves the same without it.

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,121 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+    double defaultCamFx = defaultIntrinsicsCam->ImageFocalLengthX();
+    double defaultCamFy = defaultIntrinsicsCam->ImageFocalLengthY();
+    double defaultCamCx = defaultIntrinsicsCam->ImageOpticalCentreX();
+    double defaultCamCy = defaultIntrinsicsCam->ImageOpticalCentreY();
+
+    EXPECT_GT(defaultCamWidth, 0u);
+    EXPECT_GT(defaultCamHeight, 0u);
```
jacobperron: These two checks are redundant with the `EXPECT_EQ` calls below.
```suggestion
```

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,121 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+    EXPECT_NEAR(defaultCamFx, 277.127, error);
+    EXPECT_NEAR(defaultCamFy, 277.127, error);
+    EXPECT_DOUBLE_EQ(defaultCamCx, 160);
+    EXPECT_DOUBLE_EQ(defaultCamCy, 120);
+  }
+
+  // get the 'intrinsics_camera' sensor, <intrinsics> tag is used explicitly for this sensor
+  // where the intrinsics provided are same as gazebo default.
+  std::string intrinsicsCameraName = "intrinsics_camera_sensor";
+  sensors::SensorPtr intrinsicsSensor = sensors::get_sensor(intrinsicsCameraName);
```
jacobperron:
```suggestion
  sensors::SensorPtr intrinsicsSensor = sensors::get_sensor( "intrinsics_camera_sensor");
```

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,121 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+
+    EXPECT_GT(camWidth, 0u);
+    EXPECT_GT(camHeight, 0u);
+    EXPECT_EQ(camWidth, 320u);
+    EXPECT_EQ(camHeight, 240u);
+    EXPECT_DOUBLE_EQ(camFx, 277);
+    EXPECT_DOUBLE_EQ(camFy, 277);
+    EXPECT_DOUBLE_EQ(camCx, 160);
+    EXPECT_DOUBLE_EQ(camCy, 120);
+  }
```
jacobperron:
```suggestion
  double camFx = intrinsicsCam->ImageFocalLengthX();
  double camFy = intrinsicsCam->ImageFocalLengthY();
  double camCx = intrinsicsCam->ImageOpticalCentreX();
  double camCy = intrinsicsCam->ImageOpticalCentreY();

  EXPECT_EQ(camWidth, 320u);
  EXPECT_EQ(camHeight, 240u);
  EXPECT_DOUBLE_EQ(camFx, 277);
  EXPECT_DOUBLE_EQ(camFy, 277);
  EXPECT_DOUBLE_EQ(camCx, 160);
  EXPECT_DOUBLE_EQ(camCy, 120);
```

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,121 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+
+  // connect to new frame event
```
jacobperron: I'm not sure what we're testing below this line. IIUC, we're checking that the both sensors return the same image. The opposite might actually be more interesting: demonstrating that changes to the intrinsics does in fact change the image.

gazebo/rendering/Camera.cc
```diff
@@ -1094,6 +1116,30 @@ double Camera::FarClip() const
...
+//////////////////////////////////////////////////
+double Camera::ImageOpticalCentreX() const
+{
+  return this->dataPtr->cameraIntrinsicMatrix(0, 2);
+}
+
+//////////////////////////////////////////////////
+double Camera::ImageOpticalCentreY() const
+{
+  return this->dataPtr->cameraIntrinsicMatrix(1, 3);
```
jacobperron: I _think_ the element we want is `(1, 2)`, no?

gazebo/rendering/Camera.cc
```diff
@@ -234,13 +236,15 @@ void Camera::UpdateCameraIntrinsics(
...
+  this->dataPtr->cameraIntrinsicMatrix = this->BuildIntrinsicMatrix(
+             _cameraIntrinsicsFx, _cameraIntrinsicsFy,
+             _cameraIntrinsicsCx, _cameraIntrinsicsCy);
 
   if (this->camera != nullptr)
   {
     this->camera->setCustomProjectionMatrix(true,
         Conversions::Convert(this->cameraProjectiveMatrix));
   }
-
```
jacobperron: nitpick: avoid making unrelated whitespace changes.

gazebo/rendering/Camera.cc
```diff
@@ -2135,6 +2182,28 @@ ignition::math::Matrix4d Camera::ProjectionMatrix() const
   return Conversions::ConvertIgn(this->camera->getProjectionMatrix());
 }
 
+///////////////////////////////////////////////
+void Camera::CalculateIntrinsicsFromProjectionMatrix()
+{
+  const ignition::math::Matrix4d& m = this->ProjectionMatrix();
```
jacobperron: I suggest this variable `P` or `projectionMat` for clarity.

gazebo/rendering/Camera.hh
```diff
@@ -206,6 +206,14 @@ namespace gazebo
               const double _intrinsicsS,
               const double _clipNear, const double _clipFar);
 
+      /// \brief Compute the intrinsic camera matrix, this matrix is different
+      ///        than the one used by OpenGL internally and contains the camera
+      ///        calibrated values
+      /// \return intrinsic matrix
+      private: ignition::math::Matrix3d BuildIntrinsicMatrix(
```
jacobperron: Seems odd to have a private method in the middle of the public methods. Consider moving this declaration to later in the file.

gazebo/rendering/Camera.hh
```diff
@@ -393,6 +401,22 @@ namespace gazebo
...
+      /// \return X focal length in pixels
+      public: double ImageFocalLengthX() const;
+
+      /// \brief Get the Y focal length in pixels
+      /// \return Y focal length in pixels
+      public: double ImageFocalLengthY() const;
+
+      /// \brief Get the X principal point in pixels
+      /// \return X principal point in pixels
+      public: double ImageOpticalCentreX() const;
```
jacobperron: I guess we probably want to stick to the US spelling "center" instead of the British "centre"

jacobperron:
> Should we expose the API's for ImageFocalLengthX(), ImageOpticalCentreX() etc in CameraSensor similar to ImageWidth() and ImageHeight() API's?

Maybe, I'm not sure of the scenario where this might be required. I guess it's easy to add. This is probably a better question for @scpeters

test/integration/camera_sensor.cc
deepanshubansal01: I will add a unit test for the different images as well.

gazebo/rendering/Camera.cc
deepanshubansal01: yeah right, makes me wonder how unit tests were working. I will take a look. Thanks for pointing it out.

gazebo/rendering/Camera.hh
deepanshubansal01: sounds good.

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,151 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+  double camCx = cam->ImageOpticalCenterX();
+  double camCy = cam->ImageOpticalCenterY();
+  EXPECT_EQ(camWidth, 320u);
+  EXPECT_EQ(camHeight, 240u);
+  EXPECT_DOUBLE_EQ(camFx, 320);
+  EXPECT_DOUBLE_EQ(camFy, 320);
+  EXPECT_DOUBLE_EQ(camCx, 160);
+  EXPECT_DOUBLE_EQ(camCy, 120);
+
+  // connect to new frame event
```
adityapande-1995: Could you capitalize the fist letters in this and the following comments ? `` // Connect to new frame event.``

scpeters:
> > Should we expose the API's for ImageFocalLengthX(), ImageOpticalCentreX() etc in CameraSensor similar to ImageWidth() and ImageHeight() API's?
> 
> Maybe, I'm not sure of the scenario where this might be required. I guess it's easy to add. This is probably a better question for @scpeters

I think it would be simpler to just use the `CameraSensor::Camera()` API to get a `rendering::Camera` pointer and then use these APIs added in this PR from that object. So, I don't think additional changes are needed to the CameraSensor

https://github.com/osrf/gazebo/blob/gazebo11/gazebo/sensors/CameraSensor.hh#L93

test/integration/camera_sensor.cc
scpeters: 👍

adityapande-1995 (approved):

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,151 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+  EXPECT_EQ(intrinsicsCamHeight, 240u);
+  EXPECT_DOUBLE_EQ(intrinsicsCamFx, 277);
+  EXPECT_DOUBLE_EQ(intrinsicsCamFy, 277);
+  EXPECT_DOUBLE_EQ(intrinsicsCamCx, 160);
+  EXPECT_DOUBLE_EQ(intrinsicsCamCy, 120);
+
+  // Get the 'camera' sensor, <intrinsics> tag is used explicitly for this sensor
+  // where the intrinsics provided are different from the gazebo default.
+  sensors::SensorPtr sensor = sensors::get_sensor("camera_sensor");
+  sensors::CameraSensorPtr camSensor =
```
adityapande-1995: Can we make the names of the test cameras consistent like ``camWithoutIntrisics`` , ``camDefaultIntrinsics`` and ``camCustomIntrinsics`` ? Other than that, looks good to me.

jacobperron (approved):
LGTM

I'm not sure if we should expect CI to be completely green. I'll leave that up to @scpeters

test/integration/camera_sensor.cc
```diff
@@ -2268,3 +2268,151 @@ TEST_F(CameraSensor, CheckNewAndLegacyDistortionModes)
...
+      std::dynamic_pointer_cast<sensors::CameraSensor>(sensor);
+  EXPECT_TRUE(camSensor != nullptr);
+  rendering::CameraPtr cam = camSensor->Camera();
```
adityapande-1995: Similarly for here, it would nice to make the names consistent across the 3 cameras for the ``SensorPtr`` and the ``CameraPtr`` objects.

adityapande-1995 (approved):
Looks good to me !

scpeters:
the new test is failing in macOS; I'll take a quick look and disable it if there's not an easy fix:

* https://build.osrfoundation.org/job/gazebo-ci-pr_any-homebrew-amd64/2775/testReport/junit/(root)/CameraSensor/CheckIntrinsics/

scpeters:
> the new test is failing in macOS; I'll take a quick look and disable it if there's not an easy fix:
> 
> * https://build.osrfoundation.org/job/gazebo-ci-pr_any-homebrew-amd64/2775/testReport/junit/(root)/CameraSensor/CheckIntrinsics/

disabled on macOS in https://github.com/osrf/gazebo/pull/3245/commits/71a7397cb894ea5a2845db401ddd39b7f28b573f

scpeters (approved):

PR #3249, opened 2022-08-09
CustomPSSMShadowCamera: support custom projection matrix

mogumbo:
PSSM shadows only support a regular camera matrix defined by horizontal_fov. If you instead use a custom matrix defined by camera intrinsics, the PSSM algorithm does not build a new shadow camera matrix for each of the 3 shadow maps. This PR fixes that deficiency.

Example camera using horizontal_fov:
```
           <camera>
            <horizontal_fov>0.28</horizontal_fov>
            <image>
              <width>2048</width>
              <height>2048</height>
            </image>
            <clip>
              <near>0.1</near>
              <far>500</far>
            </clip>
          </camera>
```

Example camera using intrinsics:
```
           <camera>
            <lens>
              <intrinsics>
                <fx>1434.84</fx>
                <fy>1435.36</fy>
                <cx>1010.23</cx>
                <cy>1034.08</cy>
              </intrinsics>
            </lens>
            <image>
              <width>2048</width>
              <height>2048</height>
            </image>
            <clip>
              <near>0.1</near>
              <far>500</far>
            </clip>
          </camera>
```

### DIFF
--- a/gazebo/rendering/CustomPSSMShadowCameraSetup.cc
+++ b/gazebo/rendering/CustomPSSMShadowCameraSetup.cc
@@ -465,8 +465,10 @@ void CustomPSSMShadowCameraSetup::calculateShadowMappingMatrix(
 }
 
 //////////////////////////////////////////////////
-Ogre::Matrix4 CustomPSSMShadowCameraSetup::buildSimplePerspectiveMatrix(
-    const Ogre::Real _near, const Ogre::Real _far) const
+// Build a simple perspective projection matrix using only near and far clipping
+// planes.
+static Ogre::Matrix4 buildSimplePerspectiveMatrix(
+    const Ogre::Real _near, const Ogre::Real _far)
 {
   return Ogre::Matrix4(1, 0, 0, 0,
     0, 1, 0, 0,
@@ -636,7 +638,8 @@ void CustomPSSMShadowCameraSetup::getShadowCamera(const Ogre::SceneManager *_sm,
     // elements differ in the forth or fifth digit).
     Ogre::Matrix4 oldNearFarMat = buildSimplePerspectiveMatrix(oldNear, oldFar);
     Ogre::Matrix4 newNearFarMat = buildSimplePerspectiveMatrix(nearDist, farDist);
-    cam->setCustomProjectionMatrix(true, oldCustomProjMat * oldNearFarMat.inverse() * newNearFarMat);
+    cam->setCustomProjectionMatrix(true, oldCustomProjMat *
+                                   oldNearFarMat.inverse() * newNearFarMat);
   }
 
   // Replaced LiSPSMShadowCameraSetup::getShadowCamera() with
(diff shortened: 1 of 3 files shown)


gazebo/rendering/CustomPSSMShadowCameraSetup.cc
```diff
@@ -624,6 +649,10 @@ void CustomPSSMShadowCameraSetup::getShadowCamera(const Ogre::SceneManager *_sm,
   // restore near/far
   cam->setNearClipDistance(oldNear);
   cam->setFarClipDistance(oldFar);
+  if (cam->isCustomProjectionMatrixEnabled())
+  {
+    cam->setCustomProjectionMatrix(true, oldCustomProjMat);
```
jacobperron, line 657: This looks like we're resetting the custom projection matrix back to the old matrix. So we only want to modify the custom projection matrix temporarily (before calling the clip distance methods)?

gazebo/rendering/CustomPSSMShadowCameraSetup.cc
```diff
@@ -464,6 +464,16 @@ void CustomPSSMShadowCameraSetup::calculateShadowMappingMatrix(
   }
 }
 
+//////////////////////////////////////////////////
+Ogre::Matrix4 CustomPSSMShadowCameraSetup::buildSimplePerspectiveMatrix(
```
jacobperron: Since this function is only used within this translation unit, and it does not depend on member variables, I suggest only declaring it in this .cc file as a static free function, e.g.

```cpp
static Ogre::Matrix4 buildSimplePerspectiveMatrix(const Ogre::Real _near, const Ogre::Real _far)
{
  ...
```

And then we don't need to modify the header file.

gazebo/rendering/CustomPSSMShadowCameraSetup.cc
mogumbo: Correct. In the case of non-custom projection matrices, lines 625 and 626 build a new temporary matrix and lines 650 and 651 rebuild the original matrix. Those lines have no effect on a custom matrix, so I found a different way to build a temporary matrix and restore the original.

Line 619 points out that this is all very "hacky," so the original author of this code probably would have preferred a cleaner solution but didn't want all the refactoring that it would have required.

gazebo/rendering/CustomPSSMShadowCameraSetup.cc
mogumbo: I can do that. Out of curiosity, does adding this as a method break and project rules? I don't believe it will break ABI.

scpeters: > I can do that. Out of curiosity, does adding this as a method break and project rules? I don't believe it will break ABI.

since it is not a `virtual` function, it would not break ABI to add it to the installed header file, but keeping it in the `.cc` file helps limit our API surface area

mogumbo: Done

scpeters (approved):

Issue #1476, opened 2015-03-02
Implement AddXXXForce for Bullet


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/1476)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
Related ODE [pull request #1456](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/1456)

jeduardo211:
I am thinking about working on a series of PRs to implement or correct the implementation of the methods to add and set forces and torques in order to ensure that the behavior of theses methods is consistent across all physics engines, but I would like to ask two questions before starting to work on this:

1. Would such PRs be acceptable and able to be merged in the last version of Gazebo classic (v11)?
2. Is there any tests for these methods, I cannot find them in the physics directory, can anyone point me to their location?

@chapulina Are you be able to answer these two questions or can you point me to anyone who can answer them?

scpeters:
> 1. Would such PRs be acceptable and able to be merged in the last version of Gazebo classic (v11)?

Yes! I would be happy to review these pull requests.

> 2. Is there any tests for these methods, I cannot find them in the physics directory, can anyone point me to their location?

There are related tests in [physics_link.cc](https://github.com/osrf/gazebo/blob/gazebo11/test/integration/physics_link.cc#L209-L218). If you update the logic in that linked code block you can enable the bullet tests, which should fail until the implementation is fixed

scpeters:
fixed by #3255

PR #3255, opened 2022-09-04
BulletLink: add and set force and torque

jeduardo211:
This PR adds the implementation of methods for applying forces and torques to links when using the Bullet physics engine.

The following methods are now implemented in the BulletLink class:

- AddForce
- AddRelativeForce
- AddForceAtWorldPosition
- AddForceAtRelativePosition
- AddLinkForce
- AddTorque
- AddRelativeTorque
- WorldTorque

The implementation of the above methods was designed to match ODE and DART behavior.

The WorldTorque method always returned a vector of zeros, and had a implementation for returning the totalTorque from the btRigidBody object that was commented out. I couldn't not find a reason for this in the commit history nor in the Github issues, please advise if there is a known problem with that implementation.

Additionally, the implementation of the following methods was modified to correct their behavior:

- SetForce
- SetTorque

Expectation, is that calling these methods should respect the following sequence of steps:

1. Initially, the link has world force F1 and world torque T1
2. SetForce is called with _force = F2
3. The link now has force equal to F2 and torque T1

This behavior is different from the one contained in the previous implementation, which was the following:

1. Initially, the link has world force F1 and world torque T1
2. SetForce is called with _force = F2
3. The link now has force equal to F1 + F2, and torque equal to T1

Also, the UpdateMass method was modified to also update the CoG position. This was required to ensure that the updates to CoG that exist in the AddForce test were adequately reflected in the btRigidBody object. Note that this is consistent with the implementation of the UpdateMass method of the ODELink class.

In order to test the above methods, the AddForce test of the physics_link integration tests is now enabled for Bullet, and executes successfully.

**Important:** The OnWrenchMsg test is still disabled for bullet, because I could not find a way to persist the application of forces across steps with this physics engine.

This PR solves issue #1476.

### DIFF
--- a/gazebo/physics/bullet/BulletLink.cc
+++ b/gazebo/physics/bullet/BulletLink.cc
@@ -242,6 +242,8 @@ void BulletLink::UpdateMass()
     }
     this->rigidLink->setMassProps(this->inertial->Mass(),
         BulletTypes::ConvertVector3(this->inertial->PrincipalMoments()));
+    // In case the CoG position changed
+    this->OnPoseChange();
   }
 }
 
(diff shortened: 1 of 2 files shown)


scpeters (approved):
well done! You've properly enabled the test and implemented all the APIs

thanks for the contribution!

jeduardo211:
Thanks @scpeters, I will start work on Simbody implementation of these APIs

Issue #555, opened 2013-03-05
Thread in sensor manager can potentially sleep forever


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/issues/555)) by Ian Chen (Bitbucket: [Ian Chen](https://bitbucket.org/%7Beaa6fca5-6deb-43f6-907f-971c144735dd%7D/), GitHub: [iche033](https://github.com/iche033)).**

----------------------------------------
in `SensorManager::SensorContainer::RunLoop()`

When calculating sleep time:
```
#!c++

  // Calculate an appropriate sleep time.
  if (maxUpdateRate > 0)
    sleepTime.Set(1.0 / (maxUpdateRate));
  else
    sleepTime.Set(0, 1e6);
```
if no update rate is specified then `maxUpdateRate` evaluates to `GZ_DBL_MIN` which results in `1.0/GZ_DBL_MIN` sleep time

PR #3258, opened 2022-09-10
Lunar DEMs : Make Heightmap loader use spherical coordinates.

adityapande-1995:
This PR is an ongoing set of PRs to make lunar dem support available in gazebo classic.

### For gzserver
``HeightmapDataLoader::LoadDEMAsTerrain()`` needs access to the world's spherical coordiantes object to load lunar DEMs correctly. This in turn is called by ``HeightmapDataLoader::LoadTerrain()`` , which is loaded by ``HeightmapShape::LoadTerrainFile()`` which ultimately has access to the spherical coordiantes object.

This PR overloads the above functions so that the spherical coordinates pointer is available to the DEM object.

```
# Inside gazebo/gazebo/physics/
HeightmapDataLoader::LoadDEMAsTerrain() --> Loads DEMs, needs access to SphericalCoordinates
  |
HeightmapDataLoader::LoadTerrainFile()
  |
HeightmapShape::LoadTerrain() --> Has access to SphericalCoordinates
```

### For gzclient
```
# inside gazebo/rendering

HeightmapLoader::LoadTerrainFile() --> Needs SphericalCoordiantesPtr
 | 
Heightmap::Load() --> queries spherical coordinates using a service
 |
World.cc --> Exposes service to return world surface type.
```

## Possible issues
This might still lead to some ABI incompatibility issues, and I can go for a different approach then, probably by renaming functions or deprecating old ones.

It is also possible to direct use the service in ``HeightmapDataLoader::LoadDEMAsTerrain()``, but that *might* slow down the loading of DEMs even in server mode.


gazebo/rendering/Heightmap.hh
```diff
@@ -23,6 +23,7 @@
 #include <ignition/math/Vector3.hh>
 #include <ignition/math/Vector2.hh>
 
+#include <gazebo/common/SphericalCoordinates.hh>
```
scpeters: if this isn't needed here, I would remove it or move it to the `.cc` file

gazebo/common/HeightmapData.hh
```diff
@@ -82,12 +82,33 @@ namespace gazebo
...
 
+      /// \brief Same as LoadTerrainFile with filename, but also
+      /// considers the surface type in the supplied Spherical Coordinates.
+      /// \param[in] _filename The path to the terrain file.
+      /// \param[in] _sphericalCoordinates SphericalCoordinatesPtr from the
+      /// world.
+      /// \return 0 when the operation succeeds to load a file or -1 when fails.
+      public: static HeightmapData *LoadTerrainFile(
+          const std::string &_filename,
+          common::SphericalCoordinatesPtr _sphericalCoordinates);
```
scpeters, line 99: I think this file should now include `gazebo/common/SphericalCoordinates.hh`

gazebo/common/HeightmapData.cc
```diff
@@ -95,6 +112,41 @@ HeightmapData *HeightmapDataLoader::LoadTerrainFile(
     return LoadDEMAsTerrain(_filename);
   }
 }
+
+//////////////////////////////////////////////////
+/// Load terrain files taking into account the spherical
+/// coordinates surface type.
+HeightmapData *HeightmapDataLoader::LoadTerrainFile(
+    const std::string &_filename,
+    common::SphericalCoordinatesPtr _sphericalCoordinates)
```
scpeters, line 93: this new method needs to be implemented in the `#else` block below

also, this file should include `gazebo/common/SphericalCoordinates.hh`

gazebo/common/HeightmapData.cc
```diff
@@ -95,6 +112,41 @@ HeightmapData *HeightmapDataLoader::LoadTerrainFile(
...
+  // Check if the heightmap file is an image
+  if (fileFormat == "JPEG" || fileFormat == "PNG")
+  {
+    // Load the terrain file as an image
+    return LoadImageAsTerrain(_filename);
+  }
+  else
+  {
+    // Load the terrain file as a DEM
+    return LoadDEMAsTerrain(_filename, _sphericalCoordinates);
```
scpeters, line 119: this new function is almost identical to the existing overload. I'd recommend avoiding code duplication by calling one method from the other

scpeters (commented):
I'd like to see a unit test for the new `HeightmapData::LoadTerrainFile` overload in HeightmapData_TEST.cc

I made several other comments as well that I have implemented in 9cdeebcbe72dfd8fea675cd424b3aed05ed238b2 that you may consider cherry-picking

gazebo/common/HeightmapData.cc
adityapande-1995: Hmm, wouldn't that be the section that runs when we don't have ``GDAL`` ? Since this overload is meant for the original ``LoadTerrainFile(const std::string &_filename)``, shouldn't it live in the same ``if`` block ?

gazebo/rendering/Heightmap.hh
adityapande-1995: Done https://github.com/osrf/gazebo/pull/3258/commits/fd3b266bf7e0ae96ba4597003c747a22c8a9df48

gazebo/common/HeightmapData.hh
adityapande-1995: Done https://github.com/osrf/gazebo/pull/3258/commits/fd3b266bf7e0ae96ba4597003c747a22c8a9df48

gazebo/common/HeightmapData.cc
scpeters: the header file defines [both overloads](https://github.com/osrf/gazebo/blob/fd3b266bf7e0ae96ba4597003c747a22c8a9df48/gazebo/common/HeightmapData.hh#L77-L94) regardless of whether GDAL is installed, so both overloads should be defined in the `#ifdef HAVE_GDAL` and `#else` blocks

this probably applies to `LoadDEMAsTerrain` as well, now that I think of it

I see the following related test failure on homebrew: https://build.osrfoundation.org/job/gazebo-ci-pr_any-homebrew-amd64/2817/testReport/(root)/INTEGRATION_heightmap/test_ran/

~~~
555: [ RUN      ] HeightmapTest.DartCollisionDetectorSelectionBullet
555: dyld: lazy symbol binding failed: Symbol not found: __ZN6gazebo6common19HeightmapDataLoader15LoadTerrainFileERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEEN5boost10shared_ptrINS0_20SphericalCoordinatesEEE
555:   Referenced from: /usr/local/lib/libgazebo_physics.11.dylib
555:   Expected in: flat namespace
555: 
555: dyld: Symbol not found: __ZN6gazebo6common19HeightmapDataLoader15LoadTerrainFileERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEEN5boost10shared_ptrINS0_20SphericalCoordinatesEEE
555:   Referenced from: /usr/local/lib/libgazebo_physics.11.dylib
555:   Expected in: flat namespace
555: 
555/676 Test #555: INTEGRATION_heightmap .................................Subprocess aborted***Exception:   1.38 sec
~~~

adityapande-1995: Added overloads, but for ``LoadDEMAsTerrain``, I simply added a warning, as we can't use GDAL anyway in this case : https://github.com/osrf/gazebo/pull/3258/commits/38e0689a551182299fab778f5d7e0bbe0f14ce92

gazebo/common/HeightmapData.cc
adityapande-1995: Done : https://github.com/osrf/gazebo/pull/3258/commits/26e9bd775983270523aef5517f77e2711bb1ed25

adityapande-1995:
> I'd like to see a unit test for the new `HeightmapData::LoadTerrainFile` overload in HeightmapData_TEST.cc
> 
> I made several other comments as well that I have implemented in [9cdeebc](https://github.com/osrf/gazebo/commit/9cdeebcbe72dfd8fea675cd424b3aed05ed238b2) that you may consider cherry-picking

Added test https://github.com/osrf/gazebo/pull/3258/commits/8dc627054d2297c34de55794fc63438875c188e5

gazebo/common/HeightmapData.hh
```diff
@@ -82,12 +83,33 @@ namespace gazebo
       public: static HeightmapData *LoadTerrainFile(
           const std::string &_filename);
 
+      /// \brief Same as LoadTerrainFile with filename, but also
+      /// considers the surface type in the supplied Spherical Coordinates.
+      /// \param[in] _filename The path to the terrain file.
+      /// \param[in] _sphericalCoordinates SphericalCoordinatesPtr from the
+      /// world.
+      /// \return 0 when the operation succeeds to load a file or -1 when fails.
```
scpeters: this doc-string and the `\return` in the other overload just above are incorrect. These functions return a pointer to memory allocated to store the heightmap data (and passes ownership of the data along with it) when the operation is successful or `nullptr` if it is not successfully loaded

gazebo/common/HeightmapData.hh
```diff
@@ -82,12 +83,33 @@ namespace gazebo
...
       private: static HeightmapData *LoadDEMAsTerrain(
           const std::string &_filename);
 
+      /// \brief Load a DEM specified by _filename as a terrain file,
+      /// also taking into account the surface type in the supplied Spherical
+      /// Coordiantes.
+      /// \param[in] _filename The path to the terrain file.
+      /// \param[in] _sphericalCoordinates SphericalCoordinatesPtr from the
+      /// world.
+      /// \return 0 when the operation succeeds to load a file or -1 when fails.
```
scpeters: fix doc-string per comment above

gazebo/common/HeightmapData.hh
```diff
@@ -82,12 +83,33 @@ namespace gazebo
...
+      public: static HeightmapData *LoadTerrainFile(
+          const std::string &_filename,
+          common::SphericalCoordinatesPtr _sphericalCoordinates);
+
       /// \brief Load a DEM specified by _filename as a terrain file.
       /// \param[in] _filename The path to the terrain file.
       /// \return 0 when the operation succeeds to load a file or -1 when fails.
```
scpeters: fix doc-string per comment above

gazebo/physics/World.hh
```diff
@@ -669,6 +669,12 @@ namespace gazebo
       private: bool ShadowCasterMaterialNameService(
           ignition::msgs::StringMsg &_response);
 
+      /// \brief Callback for "<this_name>/spherical_coordinates_surface_type" service.
+      /// \param[out] _response Message containing spherical ciirdinates surface name.
```
scpeters:
```suggestion
      /// \param[out] _response Message containing spherical coordinates surface type.
```

adityapande-1995: Done : https://github.com/gazebosim/gazebo-classic/pull/3258/commits/3b954aa19fbb64250466f861a69289b55ca7c853

gazebo/common/HeightmapData.hh
adityapande-1995: Done : https://github.com/gazebosim/gazebo-classic/pull/3258/commits/3b954aa19fbb64250466f861a69289b55ca7c853

gazebo/common/HeightmapData.hh
adityapande-1995: Done https://github.com/gazebosim/gazebo-classic/pull/3258/commits/3b954aa19fbb64250466f861a69289b55ca7c853

gazebo/common/HeightmapData.hh
adityapande-1995: Done in https://github.com/gazebosim/gazebo-classic/pull/3258/commits/3b954aa19fbb64250466f861a69289b55ca7c853, fixed the ``return`` statement in a couple of other places as well.

scpeters (approved):

PR #3294, opened 2023-01-30
Set initial sim time from the command-line

audrow:
This PR adds the `--initial_sim_time` flag to `gzserver`. If the flag and a value is not provided, the behavior is not changed. This works with large times, such as epoch time and floating point numbers (see the test).

FYI @scpeters

### DIFF
--- a/gazebo/Server.cc
+++ b/gazebo/Server.cc
@@ -194,6 +194,8 @@ bool Server::ParseArgs(int _argc, char **_argv)
      "Absolute path in which to store state data")
     ("record_period", po::value<double>()->default_value(-1),
      "Recording period (seconds).")
+    ("initial_sim_time", po::value<double>(),
+     "Initial simulation time (seconds).")
     ("record_filter", po::value<std::string>()->default_value(""),
      "Recording filter (supports wildcard and regular expression).")
     ("record_resources", "Recording with model meshes and materials.")
@@ -403,6 +405,23 @@ bool Server::ParseArgs(int _argc, char **_argv)
         gzerr << "Specified profile [" << profileName << "] was not found."
               << std::endl;
       }
+    }
+  }
+
+  if (this->dataPtr->vm.count("initial_sim_time"))
+  {
+    try
+    {
+      physics::get_world()->SetSimTime(
+          common::Time(this->dataPtr->vm["initial_sim_time"].as<double>()));
+      gzmsg << "Setting initial sim time to [" <<
+        physics::get_world()->SimTime() << "]\n" << std::endl;
+    }
+    catch(...)
+    {
+      gzerr << "Unable to cast initial_sim_time[" <<
+        this->dataPtr->vm["initial_sim_time"].as<double>() << "] "
+        << "to double for setting initial sim time\n" << std::endl;
     }
   }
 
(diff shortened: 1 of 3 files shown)


scpeters:
> This PR adds the `--initial_sim_time` flag to `gzserver`.

I didn't realize that our `gzserver` CLI flags use `_` as a separator. It makes sense to keep this patter, but I'm glad that gz-sim is using `--initial-sim-time` instead of `--initial_sim_time`

gazebo/Server.cc
```diff
@@ -194,6 +194,8 @@ bool Server::ParseArgs(int _argc, char **_argv)
      "Absolute path in which to store state data")
     ("record_period", po::value<double>()->default_value(-1),
      "Recording period (seconds).")
+    ("initial_sim_time", po::value<double>(),
+     "Initial simulation time (seconds).")
```
scpeters: I believe the order of these arguments determines the formatting of the help string, so I would move this out of the middle of the `record_*` arguments to be next to `iters`

test/integration/world_with_initial_sim_time_from_cli.cc
```diff
@@ -0,0 +1,65 @@
...
+
+const double initialSimTime = 1675117690.123456;
+
+/// \brief Helper class that initializes each test.
+class WorldWithInitialSimTimeFromCliTest : public ServerFixture
+{
+  /// \brief Class constructor.
+  public: WorldWithInitialSimTimeFromCliTest()
+  {
+    // Start the server with a log file and paused.
```
scpeters:
```suggestion
    // Start the server with an initial sim time and paused.
```

test/integration/world_with_initial_sim_time_from_cli.cc
```diff
@@ -0,0 +1,65 @@
...
+    this->LoadArgs("-u --initial_sim_time " + std::to_string(initialSimTime));
+    this->world = physics::get_world("default");
+
+    // Prepare the transport.
+    this->node = transport::NodePtr(new transport::Node());
+    this->node->Init();
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(300));
```
scpeters: this sleep and the `this->node` lines probably aren't needed for this test

scpeters (commented):
LGTM other than a few minor comments

gazebo/Server.cc
audrow: Fixed in https://github.com/gazebosim/gazebo-classic/pull/3294/commits/5cc4db31189963f21ce9e572427e7a03f42a6869.

test/integration/world_with_initial_sim_time_from_cli.cc
audrow: Fixed in https://github.com/gazebosim/gazebo-classic/pull/3294/commits/5078e9d623e034301cc897cc4395ebe31bd26ea0.

test/integration/world_with_initial_sim_time_from_cli.cc
audrow: I think you're right. I removed these in https://github.com/gazebosim/gazebo-classic/pull/3294/commits/5078e9d623e034301cc897cc4395ebe31bd26ea0.

gazebo/Server.cc
scpeters: thanks; I pushed it one more line in https://github.com/gazebosim/gazebo-classic/pull/3294/commits/7f872b49088d85822df50626c60b2661cf0b4c8f

scpeters (approved):
I'll merge after reviewing the CI results

peci1:
Shouldn't this PR also adjust the time that is set in `World::ResetTime()`?

https://github.com/gazebosim/gazebo-classic/blob/17e09f574a4f39caff279cd70364cd1a3ea46f70/gazebo/physics/World.cc#L1456

PR #3305, opened 2023-03-14
Allow user to name a specific light that will generate lens flare.

mogumbo:
# 🎉 New feature

Closes # not applicable

## Summary
Allow user to name a specific light that will generate lens flare.

## Checklist
- [ ] Added tests
- [ ] Added example and/or tutorial

### DIFF
--- a/gazebo/rendering/LensFlare.cc
+++ b/gazebo/rendering/LensFlare.cc
@@ -560,30 +560,22 @@ void LensFlare::Update()
 
   LightPtr light;
 
-  // Use the preferred light, if there is one
+  // Use the specified light, if there is one
   if (!this->dataPtr->lightName.empty())
   {
     light = this->dataPtr->camera->GetScene()->LightByName(
         this->dataPtr->lightName);
-    static bool notPrintedYet = true;
-    if (!light && notPrintedYet)
-    {
-      gzwarn << "There exists no light named " << this->dataPtr->lightName
-             << std::endl;
-      notPrintedYet = false;
-    }
   }
-
-  // Get the first directional light
-  if (!light)
+  else // Get the first directional light
   {
-    for (unsigned int i = 0; i < this->dataPtr->camera->GetScene()->LightCount();
-        ++i)
+    for (unsigned int i = 0;
+         i < this->dataPtr->camera->GetScene()->LightCount(); ++i)
     {
-      LightPtr lp = this->dataPtr->camera->GetScene()->LightByIndex(i);
-      if (lp->Type() == "directional")
+      LightPtr directionalLight =
+          this->dataPtr->camera->GetScene()->LightByIndex(i);
+      if (directionalLight->Type() == "directional")
       {
-        light = lp;
+        light = directionalLight;
         break;
       }
     }
--- a/gazebo/rendering/LensFlare.cc
+++ b/gazebo/rendering/LensFlare.cc
@@ -40,6 +40,9 @@ namespace gazebo
       /// \brief Pointer to light
       public: LightPtr light;
 
+      /// \brief Pointer to light
+      public: std::string lightName;
+
       /// \brief Pointer to camera
       public: CameraPtr camera;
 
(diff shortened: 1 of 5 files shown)


mogumbo:
This code doesn't work yet. It tries to select the light named by the user, but sometimes that light has not been added to the scene before the first run of LensFlare::Update(). In that case the first directional light is selected and Update() is never called again: https://github.com/gazebosim/gazebo-classic/pull/3305/files#diff-49074ecb941c9d3ea68dec38fb5d75cd9fe66f3f7a0822715dd11e4845d8189dR601

I would like to be able to be notified when the named light is added to the scene and then run Update() again, but I haven't found a way to do this.

An alternative is to only stop running Update() once the named light is found, if there is one:
```
  // disconnect
  if (this->dataPtr->lightName.empty() ||
      this->dataPtr->lightName == this->dataPtr->lightNameCurrentlyInUse)
  {
    this->dataPtr->preRenderConnection.reset();
  }
```
This introduces a bit of overhead while Update() is repeatedly running. It probably wouldn't be an issue most of the time as developers would not ordinarily name a light if they didn't plan on adding it to the scene.

scpeters:
I think the LensFlare should only use the first directional light if you haven't specified the name of a light, so it would just wait for that named light to appear. I think that would be easier?

mogumbo:
> I think the LensFlare should only use the first directional light if you haven't specified the name of a light, so it would just wait for that named light to appear. I think that would be easier?

Sounds good to me. Just implemented this and it works in my sims with either a named light and a directional light. https://github.com/gazebosim/gazebo-classic/pull/3305/commits/ca6eedcf4e99d991e2db4dc8e7178c24508345b0

@scpeters Please promote this to non-draft PR if you like the changes.

scpeters:
> @scpeters Please promote this to non-draft PR if you like the changes.

I like the changes, so I've marked this ready for review. I also added an example  demonstrating this feature in https://github.com/gazebosim/gazebo-classic/pull/3305/commits/4bf7d53ecae5aecb0bc0f1dca19bdfd922942446

gazebo/rendering/LensFlare.cc
```diff
@@ -40,6 +40,9 @@ namespace gazebo
       /// \brief Pointer to light
       public: LightPtr light;
 
+      /// \brief Pointer to light
+      public: std::string lightName;
```
scpeters: I think this variable is not used in the LensFlareCompositorListenerPrivate. If not we can remove it and the corresponding `LensFlareCompositorListener::SetLightName` method

gazebo/rendering/LensFlare.cc
```diff
@@ -86,6 +89,12 @@ namespace gazebo
       this->dataPtr->light = _light;
     }
 
+    //////////////////////////////////////////////////
+    void LensFlareCompositorListener::SetLightName(std::string _name)
```
scpeters: I think this method isn't needed

gazebo/rendering/LensFlare.hh
```diff
@@ -47,10 +47,14 @@ namespace gazebo
...
 
-      /// \brief Set directional light that generates lens flare
-      /// \param[in] _light Pointer to directional light
+      /// \brief Set light that generates lens flare
+      /// \param[in] _light Pointer to light
       public: void SetLight(LightPtr _light);
 
+      /// \brief Set the name of light that generates lens flare
+      /// \param[in] _name Light that generates lens flare
+      public: void SetLightName(std::string _name);
```
scpeters: I think this can be removed

gazebo/rendering/LensFlare.cc
mogumbo: fixed

gazebo/rendering/LensFlare.cc
mogumbo: fixed

gazebo/rendering/LensFlare.hh
mogumbo: fixed

scpeters (approved):

PR #3342, opened 2023-08-31
Wheel slip nonlinear effect (breaks ABI, not for merging)

scpeters:
# 🎉 New feature

Adds a nonlinear effect to the wheel slip model (not for merging due to ABI breaks)

## Summary

Driving on deformable terrain can currently be modeled using the `ode` physics engine in Gazebo using the `WheelSlipPlugin` to get a linear slip model (steady wheel slip roughly proportional to the slope angle when driving uphill or downhill) and the plowing parameters added in https://github.com/gazebosim/gazebo-classic/pull/3164. In cases for which the linear slip model only matches experimental data within a particular range of slope angles, a nonlinear aspect to the model would be useful.

This pull request adds a nonlinear wheel slip effect in ODEPhysics in the portion of the contact callback function where the wheel plowing model is currently implemented. This nonlinear effect iterates over each contact point/normal pair and computes for each pair an equivalent slope angle from the projection of gravity and some other body forces into the wheel's longitudinal/normal plane. For each pair of collisions in contact, the equivalent slope angle is averaged over the set of contact points for that collision pair. If the equivalent slope is below the lower threshold or above the upper threshold, the longitudinal slip compliance is altered in proportion to the difference from the threshold according to the following equation:

~~~
if (aboveThreshold) {
  longitudinal_slip_compliance *=
    (1 + nonlinearSlipUpperPerDegree * degreesAboveThreshold);
}
else if (belowThreshold) {
  longitudinal_slip_compliance *=
    (1 + nonlinearSlipLowerPerDegree * degreesBelowThreshold);
}
~~~

The slope threshold values are specified in degrees in the `//nonlinear_slip/lower/degree` and `//nonlinear_slip/upper/degree` parameters with the proportionality parameters are specified in `//nonlinear_slip/lower/per_degree` and `//nonlinear_slip/upper/per_degree`

~~~
<nonlinear_slip>
  <lower>
    <degree>-6</degree>
    <per_degree>0.01</per_degree>
  </lower>
  <upper>
    <degree>6</degree>
    <per_degree>0.01</per_degree>
  </upper>
</nonlinear_slip>
~~~

## Test it

1. In a terminal, append `test/models` from the local clone of the gazebo-classic source repository to the `GAZEBO_MODEL_PATH` variable
2. Run the example world with

~~~
gazebo --verbose test/worlds/plowing_nonlinear_slip_tricycle_demo.world
~~~

The loads a world with two tricycles driving "uphill" against an inclined gravity vector on a flat plane. The inclination angle of the gravity vector is larger than the upper nonlinear slope threshold, so the nonlinear effect causes that tricycle to reach a reduced linear speed compared to the other tricycle with no nonlinear slip parameters. Reducing the gravity inclination below the threshold (such as with gravity of `-1 0 -9.75`) will cause both tricycles to drive at the same speed, since their slip parameters will be identical.

## Checklist
- [ ] Added tests
- [ ] Added example and/or tutorial

### DIFF
--- a/gazebo/physics/Model.cc
+++ b/gazebo/physics/Model.cc
@@ -297,6 +297,14 @@ void Model::LoadJoints()
 //////////////////////////////////////////////////
 void Model::Init()
 {
+  // Cache the total mass of the model
+  this->mass = 0.0;
+  for (uint64_t i = 0; i < this->modelSDFDom->LinkCount(); ++i)
+  {
+    this->mass +=
+        this->modelSDFDom->LinkByIndex(i)->Inertial().MassMatrix().Mass();
+  }
+
   // Record the model's initial pose (for reseting)
   this->SetInitialRelativePose(this->SDFPoseRelativeToParent());
   this->SetRelativePose(this->SDFPoseRelativeToParent());
@@ -1861,6 +1869,7 @@ void Model::PluginInfo(const common::URI &_pluginUri,
       << std::endl;
 }
 
+//////////////////////////////////////////////////
 std::optional<sdf::SemanticPose> Model::SDFSemanticPose() const
 {
   if (nullptr != this->modelSDFDom)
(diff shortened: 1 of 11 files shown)


scpeters:
this pull request is provided for reference, but it breaks ABI, so it is not intended for merging to gazebo11

PR #3350, opened 2023-10-04
Visual::SetPose performance improvement / minor fixes

jmachowinski:
## Summary
This commit adds caching for the SDF 'pose' element in the visual. For us this gave a 10x speed improvement in
the poseMsgs block in Scene.cc reducing the runtime from over 10ms to ~1ms.

I am not sure if the caching is allowed for all cases, as it was implemented now.

### DIFF
--- a/gazebo/rendering/Scene.cc
+++ b/gazebo/rendering/Scene.cc
@@ -2142,8 +2142,7 @@ void Scene::PreRender()
           ignition::math::Pose3d pose = msgs::ConvertIgn(pIter->second);
           GZ_ASSERT(iter->second, "Visual pointer is NULL");
           iter->second->SetPose(pose);
-          PoseMsgs_M::iterator prev = pIter++;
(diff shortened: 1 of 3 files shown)


scpeters (approved):
both of these commits make sense to me, though I suspect there's still more of a performance improvement that can be achieved in `Scene::PreRender` by preventing `Visual::SetPose` from touching any `sdf::ElementPtr` values at all, but this would break any code that reads from `dataPtr->sdf` and assumes it has the current pose. I think we can merge this as is since it is already an improvement.

Thanks for your contribution!

roncapat:
@scpeters sorry for the OT, but is there any plan for a new point release anytime soon?

scpeters:
> @scpeters sorry for the OT, but is there any plan for a new point release anytime soon?

https://github.com/gazebosim/gazebo-classic/pull/3351

